Issue #15, opened 2018-07-17
Review of starcheck evaluation of catalogs


taldcroft:
### APR3018A
- [x] 20603: idx=11 is a 10.503 mag star that gets a 160" box despite 4 other 160s.  Seems like smaller would be better.  *Looking at detailed report this looks odd and definitely needs more investigation.*  **Fixed by #32**.

### JUL0218A
- No issues

### JUN0418A
- 49214: Fewer than 3 ACQ stars with 160 boxes ==> **OK**
- [x] 49211: WARNING: [10] Search Box Size. Star too faint for > 120 box => **OK** but I wonder why a smaller box isn't better here.  Same 20433 in MAY0718A.  **OK**

### JUN1118A
- [x] 49199: two overlapping stars / boxes. Probably OK (best possible) but review.  This is actually a very good example of making the best of a difficult field. https://icxc.cfa.harvard.edu/aspect/analysis/proseco/test_loads/JUN1118A/starcheck.html#obsid49199
- [x] 20351: 140 arcsec search box for 10.268 mag star seems non-ideal.  **OK**

### JUN2318A
- 21122: nice example of teeny box right on the edge.

### MAY0718A
- [x] 19403: [10] is 60 arcsec presumably because of the edge.  Same story in 19653, 20181.  Not clear if the probability model is doing the right thing here.  Fraction of box off the edge needs to account for `manvr_err`.  Fixed by #18.

### MAY1418A
- [x] 49252, 49247, 49242: significantly overlapping search boxes with basically identical mags.  50/50 that we get neither (swapped boxes)?  There are no better stars, SAUSAGE went with 60" box on one and a hail mary 10.5 mag star.  **Review**: proseco is doing the best possible here, and correctly assigns these stars with basically 50% on each.

### MAY2118A
- No new issues.

jeanconn:
Several catalogs have Angle Too Large warnings that suggest that the CCD boundaries should be confirmed if we want to push stuff out to the edge.  I opened a new issue #17 for this as well, though I'm not sure about the best way to actually check the useful edges.

jeanconn:
It's a known issue, but we should probably also track the 

  [ ] Fid light in search box

issue somewhere as something that will need to be double-checked on the Matlab side.

taldcroft:
Closing, will open a new one for final validation testing.

PR #18, opened 2018-07-24
Fix the calculation of probability on CCD (near CCD edge)

taldcroft:
As first seen in #15 and noted in other discussions, the calculation in `calc_p_in_box` was conceptually wrong by using the search box size instead of the maneuver error + dither size.  This fixes that.  Unfortunately there is no star catalog regression test that highlights this (and for silly technical reasons it ended up being a pain to inject one into the history), so I just present the catalog diffs for 21122.  This shows the expected change of two bright stars that had their acq search boxes reduced (incorrectly) because of proximity to the CCD edge.

This does pass the `test_calc_p_on_ccd` test, but fails others (to be fixed).

**master**
```
In [2]: acqs = get_acq_catalog(21122)
In [3]: acqs['id', 'row', 'col', 'mag', 'halfw']
Out[3]: 
<AcqTable length=8>
    id      row     col     mag   halfw
  int32   float64 float64 float32 int64
--------- ------- ------- ------- -----
198708304 -134.14 -141.41    8.16   160
198705968 -129.15  179.35    9.53   160
198707824 -338.03 -132.92    9.32   160
198713024 -477.15 -249.63    7.58   160
198837864  157.15  162.82    9.58   160
198705568 -486.11  -42.52    9.70   160
267653496  490.37  -40.96    9.06    80
198708256 -357.18  502.34    8.11    60
```
**With this PR**
```
In [4]: acqs['id', 'row', 'col', 'mag', 'halfw']
Out[4]: 
<AcqTable length=8>
    id      row     col     mag   halfw
  int32   float64 float64 float32 int64
--------- ------- ------- ------- -----
198708304 -134.14 -141.41    8.16   160
198705968 -129.15  179.35    9.53   160
198707824 -338.03 -132.92    9.32   160
198837864  157.15  162.82    9.58   160
198713024 -477.15 -249.63    7.58   160
198705568 -486.11  -42.52    9.70   160
267653496  490.37  -40.96    9.06   160
198708256 -357.18  502.34    8.11   160
```

### DIFF
--- a/proseco/acq.py
+++ b/proseco/acq.py
@@ -602,39 +602,33 @@ def broadcast_arrays(*args):
     return outs
 
 
-def calc_p_in_box(row, col, box_sizes):
+def calc_p_in_box(row, col, box_size):
     """
-    Calculate the probability that star is in the search box.
+    Calculate the probability that star is in the box.
 
     This uses a simplistic calculation which assumes that ``p_in_box`` is
-    just the fraction of search box area that is within the usable portion
+    just the fraction of box area that is within the usable portion
     of the CCD, regardless of ``man_err``.
     """
-    p_in_boxes = []
-    is_scalar, box_sizes = broadcast_arrays(np.array(box_sizes))
+    p_in_box = 1.0
+    half_width = box_size / 5  # half width of box in pixels
+    full_width = half_width * 2
+    for rc, max_rc in ((row, CHAR.max_ccd_row),
+                       (col, CHAR.max_ccd_col)):
 
-    for box_size in box_sizes:
-        p_in_box = 1.0
-        half_width = box_size / 5  # half width of box in pixels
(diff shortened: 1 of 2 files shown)


taldcroft:
BTW, the real functional change for this PR is fully in cb63f60.

taldcroft:
Passing tests again.

taldcroft:
Proseco will select a bright star that is nominally off the CCD if p_acq is bigger than other available stars.  Basically you can have a roughly 50/50 that the maneuver error will bring it on to the CCD.

So I think the starcheck Angle Too Large check needs modification, maybe just turned into a yellow warning for acq.  It could be orange if the star is more than 60 arcsec + dither off the CCD, which means that proseco has a bug because then p_acq should really be very close to zero.

PR #21, opened 2018-08-01
Make HTML report (with logging and serialization)

taldcroft:
This is not entirely finished, but is a good bit of the way to the final diagnostic HTML report that will be visible in SAUSAGE or starcheck.

Along the way it improved:
- Capability to fully serialize the acquisition table and all meta data to YAML
- Event logging mechanism


taldcroft:
Not perfect but good enough for now.

PR #23, opened 2018-08-13
WIP: Guide star selection

jeanconn:
I planned to baseline aca_lts_eval version and was planning to put functional changes / improvements in separate PRs, but I was feeling pressed for time to so everything is kinda getting glopped together.  I'll track functional changes here for now and we can review

- [ ] modified bad pixel list to just be bad pixels
- [ ] added prototype hot pixel avoidance just using brightest pixel in likely pixel readout area and a threshold in the selection stages (needs a lot of work)
- [ ] modified column spoiler exclusion to match limits in starcheck warning
- [ ] set to use asymmetric dither if provided
- [ ] changed bright mag limit to 5.9 from 5.8
- [ ] it looks like minisausage was already set to ignore the part of the magnitude spoiler check that rejected all stars with a spoiler within 7 pixels (any magnitude spoiler).  The rest of the 'direct catalog search' could still use revision.

So, Action Items:

- [x] determine analysis needed to go with a new hot pixel strategy 
- [ ] determine analysis needed to either refine 'direct catalog search' or replace it

At some level, these are the same:  how do we best get a rough estimate how much another star or the local dark current will likely spoil a guide star without getting computationally crazy?

Other actions:

 - [ ] Define testing scope (annie evaluation of catalogs etc).
 - [ ] Define "desirements" for debug/info data returned to SAUSAGE


jeanconn:
As a start, I've copied in the Guide parts of mini_sausage and included the relevant bits of the SAUSAGE guide star characteristics.

proseco/gui_char.py
```diff
@@ -0,0 +1,1629 @@
...
+                        },
+                    "Pixels": {
+                        "ZPixLim": [-512.5,511.5],
+                        "YPixLim": [-512.5,511.5],
+                        "Center": [-0.5,-0.5],
+                        "EdgeBuffer": 5,
+                        "BadPixels": [
+                            [-245,0,454,454],
+                            [-511,-511,9,9],
+                            [-510,-510,-289,-289],
```
taldcroft: Should chop this down to actual bad pixels (i.e. no hot pixels in this list), and dynamically generate hot pixel list from 1 or more dark maps.  This would be from dark maps taken from before the obs date.  The "or more" option refers to dynamically replicating the current hot pixel algorithm using the 3 readouts before the date.

Eventually the whole hot pixel list will get scrapped, but in the meantime a dynamic hot pixel list will provide continuity with the existing algorithm.

proseco/gui.py
```diff
@@ -0,0 +1,358 @@
+import os
```
taldcroft: Coding and license.

proseco/gui.py
```diff
@@ -0,0 +1,358 @@
...
+from itertools import count
+
+import agasc
+import numpy as np
+from astropy.coordinates import SkyCoord, search_around_sky
+import astropy.units as u
+from Quaternion import Quat
+from Ska.quatutil import radec2yagzag
+import chandra_aca
+from . gui_char import CHAR
```
taldcroft: Should be `from .gui_char import CHAR`.

proseco/gui.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+PIX_2_ARC = STAR_CHAR["General"]["Pix2Arc"]
+ARC_2_PIX = 1.0 / PIX_2_ARC
+RAD_2_PIX = 180/np.pi*3600*ARC_2_PIX
+
+
+DITHER = 8
+MANVR_ERROR = 60
+fieldErrorPad = 0
```
taldcroft: Use all caps for module global.

proseco/gui.py
```diff
@@ -0,0 +1,358 @@
...
+    randerr = opt['Inertial']['MagErrRand']
+    magOneSigError = np.sqrt(randerr*randerr + caterr*caterr)
+    return magOneSigError, magOneSigError**2
+
+
+def select_stage_stars(ra, dec, roll, cone_stars):
+
+    stype = 'Guide'
+    opt = STAR_CHAR[stype][0]
+    opt['Type'] = stype
```
taldcroft: It seems like `opt['Type']` is hardwired to `Guide` in this module, so can that be factored out entirely for simplicity?

proseco/gui.py
```diff
@@ -0,0 +1,358 @@
...
+            stage  = check_stage(cone_stars,
+                                 not_bad & ~(cone_stars['{}_stage'.format(stype)] != -1),
+                                 stage_char, "{}_{}".format(stype, str(idx)))
+            cone_stars['{}_stage'.format(stype)][stage] = idx
+    selected = cone_stars[cone_stars['{}_stage'.format(stype)] != -1]
+    return selected
+
+
+
+def select_guide_stars(ra, dec, roll, dither=8, n=5, cone_stars=None, date=None):
```
taldcroft: Just use `stars` everywhere instead of `cone_stars`.  Then in the line below you can filter the stars to be in the ACA FOV box (with margin as necessary) to avoid useless stars in the list downstream.  This is consistent with usage in `acq.py`.  In fact it might make sense to factor out `acq.get_stars()` to a `utils.py` module and then use it here.

taldcroft:
Overall this needs function doc strings and more code comments before I can really start to review it.  So far I just picked off a few surface nits. This is also a good time to work on your code style for my benefit:

- Break logically connected lines of code into sub-blocks, separated by a single blank line and sometimes preceded by a comment.  Typically you have up to around 10 lines of code that are performing one sub-task, so the code is much easier to read if these are separated by a blank line.
- Capitalize the first word of code comments.

jeanconn:
> Overall this needs function doc strings and more code comments before I can really start to review it. 

I put this up as WIP without expecting any review just yet, so that's fine.

proseco/gui.py
jeanconn: Yes.  Indeed.  For the initial import I cut the "Acq" pieces and hardcoded type, but didn't factor out the type logic.

proseco/gui.py
jeanconn: "In fact it might make sense to factor out acq.get_stars() to a utils.py module and then use it here."

I was going to open that as an issue/suggestion that they share their star fetching etc, but then I figured you'd give me a hard time for premature optimization.

proseco/gui_char.py
jeanconn: Thought not sure on timetable for eventually coming back to eliminate the stage-based star selection to know how much time/effort to allocate to moderate improvements in this code.

taldcroft: I don't think you need to worry about matching SAUSAGE.  My understanding was that you had done reasonable matching validation for ACA LTS.  But more to the point, Sabina already agreed that we should just push forward with any obvious improvements now given that there will be significant testing of the output catalogs in conjunction with acq star selection.

taldcroft: I'm not sure about timetable either, except that it will be Pi times longer than any reasonable expectation.  I have a feeling that if you can make some low-hanging fruit improvements then this may be good enough for a while and allow me to focus (with @malgosias) on patch development.

taldcroft: I would say to start with the baby step because it is very easy and requires only a small new function to dynamically add to the bad pixels list.

The imposter-like approach is a fundamental change to the algorithm.  Remember also that currently there may be as much a problem from warm/hot pixels in the 8 background pixels as from centroid disturbances, and those issues cannot be easily detected without doing annie-like tracking simulation.

taldcroft:
About the WIP-ness, understood.  I do appreciate you putting up code at this stage so we can start a good discussion like this, so I'll work on keep my comments in scope for a WIP.

proseco/gui_char.py
jeanconn: Right.  I will definitely do the baby step.  If we have time/inclination for more we can look at more steps.

I think that allowing us to select 6.0 mag stars that are presently being excluding due to the presence of 2000 e-/sec hot pixels would be a step in the right direction even if the selection would not be robust to all problems (such as background issues).  The current algorithm is also not robust to all problems.

jeanconn: Regarding "I don't think you need to worry about matching SAUSAGE."

Agreed.  I just thought for my process that it made a lot of sense to baseline the "as close to SAUSAGE as we're going to get" that is what the code is basically already doing, and then have discrete improvements from that baseline in new PRs against this branch.

taldcroft: But getting close to SAUSAGE at any point in this code is irrelevant since final testing will be completely independent of the original SAUSAGE selections (except to demonstrate that the new catalogs are better by some metric).  So you will spend time testing now (validating that it gets close to SAUSAGE), for no impact later.

taldcroft: And yes, the dynamic hot pixel list should be mag dependent.

taldcroft:
Just vaguely lurking right now, but :+1: on 46eeb0a0.  I was thinking whhaaaa first time seeing that file.

jeanconn:
Thanks! Yes, my plan was to move everything that we actually use out of that from-matlab characteristics structure anyway, but I figured I'd start by simplifying in place and try to work so that everything stayed basically functional.

taldcroft:
Holy frijoles!

jeanconn:
Just lots of little commits while taking this apart and putting it back together, while trying to remember all the little oddities.

proseco/gui.py
```diff
@@ -0,0 +1,375 @@
...
+            for cc in range(row[2], row[3] + 1):
+                bp_row.append(rr)
+                bp_col.append(cc)
+    bp_row = np.array(bp_row)
+    bp_col = np.array(bp_col)
+
+    # Then for the pixel region of each candidate, see if there is a bad
+    # pixel in the region.
+    spoiled = np.zeros_like(not_bad)
+    row_extent = 4 + dither[0] * ARC_2_PIX
```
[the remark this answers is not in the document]
taldcroft: Being conservative (but not hugely conservative), the one-shot error could be 80", which means a box with halfw = 80 + dither.  With that basically there will be difficulty choosing guide stars.

We can instead stipulate that the checking here relates to the stable guide phase where the attitude error is zeroed-out.  The assumption (a good one) is that the journey while correcting the maneuver error may hit some bumps on the road (hot/bad pixels) but not enough to prevent getting to the destination.

One thing to keep in mind is that for proseco 4.0 (initial release) the guide selection here does not need to be perfect.  The intent was to just fix a few low-hanging fruit from SAUSAGE and clean the code.  Does SAUSAGE worry about bad pixels in the maneuver error box?

proseco/gui.py
```diff
@@ -0,0 +1,375 @@
...
+
+import agasc
+from Quaternion import Quat
+from Ska.quatutil import radec2yagzag
+import chandra_aca
+from mica.archive.aca_dark import get_dark_cal_image
+
+from .gui_char import CHAR, CCD
+
+# Ignore known numexpr.necompiler and table.conditions warning
```
taldcroft: Is this needed in Ska3?

proseco/gui.py
```diff
@@ -0,0 +1,375 @@
...
+warnings.filterwarnings(
+    'ignore',
+    message="using `oa_ndim == 0` when `op_axes` is NULL is deprecated.*",
+    category=DeprecationWarning)
+
+
+STAR_CHAR = CHAR["Stars"]
+PIX_2_ARC = STAR_CHAR["General"]["Pix2Arc"]
+ARC_2_PIX = 1.0 / PIX_2_ARC
+RAD_2_PIX = 180/np.pi*3600*ARC_2_PIX
```
taldcroft: Are you running PEP8 in your editor?

proseco/gui.py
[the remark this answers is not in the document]
jeanconn: SAUSAGE does not worry about bad pixels in the maneuver error box.

And regarding "low-hanging fruit", the item on my list was "do something more reasonable for bad pixel checking".  *Considering* the possibility of initial error in pixel space seemed reasonable (which we've done here), and actually implementing a secondary preference to avoid bright pixels in a bigger region around the anticipated guide star region (via stages) would be trivial if we decided to go that way.

Appropriate testing, however, is difficult to define either way (with the simplest changes or with more complicated changes to the stages).

proseco/gui.py
jeanconn: Yes, and this doesn't actually flag in my editor.  I forget where this was copied from.

proseco/gui.py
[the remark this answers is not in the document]
taldcroft: I thought the "more reasonable" ended at mag-dependence for the existing check.

But certainly giving thought to this is a good thing.  My problem is always getting stuck with the whole stage formalism when you get to questions like what is preferred:
- 9.1 mag star with no nearby bad pix but a very bright pixel 58 arcsec away
- 9.7 mag star with no bad pix within 60 + 8 arcsec
- 9.0 mag star with an 10.5 mag bad pix imposter 25 arcsec away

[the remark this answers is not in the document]
jeanconn: Well, in the other PR we've expanded to using the 2x2 pixel bin instead of the max pixel in the region (and I still have questions on that).

And regarding stanges, indeed, trivial to implement in code, non-trivial to decide upon!

PR #35, opened 2018-08-21
Use binned 2x2 pixel region to look for hot pixels that might spoil a guide star

jeanconn:
Instead of using the max pixel value in the pixel region for the guide star, use the max of the 2x2 sum in the region.


jeanconn:
Should this also check the max of the 2x2 bins that start off-by-one from the edge of the current pixel region?

jeanconn:
Needs more test cases.

PR #37, opened 2018-08-23
Fix catalog regression tests

taldcroft:
Somehow these tests were passing, and then not.  This may have been related to a local misconfiguration of my Ska3 environment on one platform.


jeanconn:
I haven't spun this up in a session, but do you need to pass the star catalog date to 

https://github.com/sot/proseco/blob/b1945ce8c438dc65cb1e694b3e6b04c45206682e/proseco/acq.py#L1086

?  I'm not that familiar with this code yet.

For me, I find all of my "these tests were passing" issues end up being due to stars actually moving on me.

jeanconn:
Still looks to me like star 37882416 has proper motion that results in a change of about 0.01 arcsecs change in yang per 30 days...

taldcroft:
Fixed by #40.

PR #48, opened 2018-08-28
Compute fid positions

taldcroft:
This is just a translation of the same Matlab function.  Testing is somewhat slim, but given heritage should be sufficient.  Downstream validation of full schedules with starcheck will also provide more testing.


jeanconn:
Gotcha.  I wasn't sure if we'd use the Matlab source or work from https://github.com/sot/starcheck/blob/master/starcheck/src/lib/Ska/Starcheck/Obsid.pm#L1752 (which I assume has to be complete for this to check the positions).

taldcroft:
I explicitly went with Matlab to avoid using (effectively) the same code to generate and check.  Plus the heritage is easier to justify in this situation.

BTW eventually bits like this and the `ACACatalogTable` could migrate out to `chandra_aca` as being generally useful, but not now.

PR #49, opened 2018-08-28
WIP on fid light selection

taldcroft:


proseco/fid.py
```diff
@@ -1,9 +1,335 @@
...
+from chandra_aca.transform import yagzag_to_pixels
+
 from . import characteristics_fid as FID
+from . import characteristics as ACQ
+
+from .core import ACACatalogTable
+
+
+def get_fid_catalog(*, detector=None, focus_offset=0, sim_offset=0,
+                    acqs=None, stars=None, dither=None,
```
jeanconn: From an API perspective, we probably want some master 'get_catalog' method that will do the acq, fid, and guide star selection and call this.  That way we won't need to try to pass the acqs back to this method from matlab (either staying on the Python side or crossing back and forth into matlab).

taldcroft: Yep, I had assumed something like that.

jeanconn: Also, what's your vision for the args that are getting slurped up in the starting asterisk?  And what's "data"?

taldcroft: No slurping, that is keyword-only arguments (Py3 new feature!).
```
In [1]: def func(*, val=2):
   ...:     print(val)
   ...:     

In [2]: func(3)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-22e29d8e19f1> in <module>()
----> 1 func(3)

TypeError: func() takes 0 positional arguments but 1 was given
```
`data` is the data arg for a Table (if that is what you are referring to).

jeanconn: Yes, I saw that data was the data arg for a Table, I just wasn't sure the use case for passing it to init (you already have a table of fids and you want to make them a FidTable?) but maybe I'm still misunderstanding.

taldcroft: Just making it a basically-functioning `Table` object and being a good citizen in subclassing (though pylint isn't quite happy).  It does get used directly when creating `cand_fids`, and also indirectly in many ways.

PR #52, opened 2018-09-01
Reboot guide selection starting from GuideTable

jeanconn:
Add guide star selection

This supersedes #23 and #35


proseco/guide.py
```diff
@@ -189,20 +185,21 @@ def search_stage(self, stage):
...
-        cand_guis[scol][reg_spoil] += GUIDE_CHAR.errs['spoiler (frac)']
-        ok = ok & ~bg_spoil & ~reg_spoil
-
         # Check for 'direct catalog search' spoilers
         mag_spoil = check_mag_spoilers(cand_guis, ok, stars, n_sigma)
         cand_guis[scol][mag_spoil] += GUIDE_CHAR.errs['spoiler (line)']
         ok = ok & ~mag_spoil
 
+        # Check for star spoilers (by light) background and edge
+        if stage['ASPQ1Lim'] > 0:
```
jeanconn: The tests that "image" stars onto an ACAImage and check for photons on the 8x8 or on the background pixels are not super speedy.  Does it make sense to only do those checks in stages where we've opened up the filter to allow nonzero ASPQ1?  Within the check, I've set it to only do the  check on stars that *also* have nonzero ASPQ1.

proseco/guide.py
```diff
@@ -306,9 +303,11 @@ def check_spoil_contrib(cand_stars, ok, stars, regfrac, bgthresh):
     reg_spoiled = np.zeros_like(ok)
     bgpix = CCD['bgpix']
     for cand in cand_stars[ok]:
+        if cand['ASPQ1'] == 0:
+            continue
         pix_dist = np.sqrt(((cand['row'] - stars['row']) **2) + ((cand['col'] - stars['col']) ** 2))
-        spoilers = pix_dist < 16
-
+        spoilers = ((np.abs(cand['row'] - stars['row']) < 9)
```
jeanconn: These limits should be in characteristics

proseco/guide.py
```diff
@@ -364,24 +364,27 @@ def check_mag_spoilers(cand_stars, ok, stars, n_sigma):
         return np.zeros(len(ok)).astype(bool)
 
     mag_spoiled = np.zeros(len(ok)).astype(bool)
-    # Search the sky around cand stars to see if they are spoiled by other stars
-    coords = SkyCoord(stars['ra'], stars['dec'], unit='deg')
```
jeanconn: In my benching, looking for star spoilers via SkyCoord seemed pretty slow and makes the code harder to read.  I think the loop that now replaces it will be a little slower in very dense fields.

proseco/guide.py
```diff
@@ -0,0 +1,551 @@
...
+        pixmax = max(np.max(bin2x2(pix)),
+                     np.max(bin2x2(pix[1:-1])),
+                     np.max(bin2x2(pix[:,1:-1])),
+                     np.max(bin2x2(pix[1:-1,1:-1])))
+        pixmax_mag = count_rate_to_mag(pixmax)
+        pixmags.append(pixmax_mag)
+    return np.array(pixmags)
+
+
+def get_pixmag_for_offset(cand_mag, offset):
```
jeanconn: Instead of setting the limit for a star's "spoiler pixel" to be a magnitude (if the star has a pixel in its region over a pseudo magnitude threshold it would be excluded), I worked with the idea that the limit is a centroid offset and then did the ~worst case calculation that should give the magnitude *for* that offset.  I don't think this is crazy, but it is different from the other limits in the characteristics.

proseco/guide.py
```diff
@@ -0,0 +1,551 @@
...
+        offchip = ((stars['row'] > CCD['row_max'])
+                   | (stars['row'] < CCD['row_min'])
+                   | (stars['col'] > CCD['col_max'])
+                   | (stars['col'] < CCD['col_min']))
+        stars['offchip'] = offchip
+
+        # Add a filter for stars that are too close to the chip edge including dither
+        r_dith_pad = self.meta['dither'][0] * GUIDE_CHAR.ARC_2_PIX
+        c_dith_pad = self.meta['dither'][1] * GUIDE_CHAR.ARC_2_PIX
+        row_min = CCD['row_min'] + (CCD['row_pad'] + CCD['window_pad'] + r_dith_pad)
```
taldcroft: The CCD is square and these checks are symmetric, so one could write:
```
row_max = CCD['max_pix'] - CCD['row_pad'] - CCD['window_pad'] - r_dith_pad
(same  for col)
out_of_bounds = (np.abs(stars['row']) < row_max) | (np.abs(stars['col']) < col_max)
```

proseco/characteristics_guide.py
```diff
@@ -0,0 +1,107 @@
+import numpy as np
+
+CCD = {'row_min': -512.5,
```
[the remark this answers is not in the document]
jeanconn: I *think* I've used 'edge' everywhere and the only bug was here at these limits, but will review.  Thanks!

proseco/characteristics_guide.py
```diff
@@ -0,0 +1,107 @@
...
+       'row_max': 511.5,
+       'col_min': -512.5,
+       'col_max': 511.5,
+       'window_pad': 7,
+       'row_pad': 8,
+       'col_pad': 1,
+       'bgpix': ['A1', 'B1', 'G1', 'H1', 'I4', 'J4', 'O4', 'P4']}
+
+PIX_2_ARC = 4.9628899999999998
```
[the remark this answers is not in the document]
jeanconn: Did you decide about how to handle this on the `proseco.acq` side?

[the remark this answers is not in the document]
taldcroft: I'm using 5 (5.0).  It shows up in two places converting a box size in arcsec to pixels.  For sizes less than 300" if you are rounding to an integer pixel then it gives the same answer.

For acq it basically does not matter since the probabilistic approach has fuzz built in, so being 0.1 pixels off in some margin calculation is not a problem.

Note that this level of precision is not actually justified since it changes by much more than that over the FOV.

proseco/guide.py
```diff
@@ -0,0 +1,551 @@
...
+                sel = cand_guis['stage'] == -1
+                cand_guis['stage'][stage_ok & sel] = idx
+            else:
+                self.log(f'Quitting stages at {idx - 1} already have {already_selected} stars')
+                break
+        self.log('Done with search stages')
+        selected = cand_guis[cand_guis['stage'] != -1]
+        selected.sort(['stage', 'MAG_ACA'])
+        if len(selected) >= ncand:
+            return selected[0:ncand]
```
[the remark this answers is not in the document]
jeanconn: Candidates are ordered two lines before this, so first by stage then by mag.

PR #63, opened 2018-09-07
Add StarsTable class

taldcroft:
This turns the `get_stars` function into a class with class-method initializers.

The real driver is to make it easy to fudge with star catalogs, and in particular generate synthetic catalogs with carefully placed spoilers etc.  Unfortunately I didn't get to that last bit yet, but at least this refactor is working and the API is specified.


jeanconn:
Do I need to check out this branch and update test_guide.py for my current faked catalogs?  Or will everything still work if stars is supplied as a Table?

taldcroft:
This is passing tests in master so I assume everything still works.  But we could just merge this and then address clean-up and subsequent functionality in separate PRs.

PR #59, opened 2018-09-06
Fid and acq concurrent optimization

taldcroft:
Optimize fid and acquisition catalogs concurrently.

Closes #43.

PR #64, opened 2018-09-07
Start a get_aca_catalog function

jeanconn:
I had something more complex but let's just start with this and work at refining requirements.


jeanconn:
If we need this to return the starcheck-style catalog and want 'size' defined, we'll need to either take that (or something or/er specific) as a keyword argument or require obsid and infer.  I think inferring from n_fids would be a little weak, but am not sure.

jeanconn:
If we want some cache of catalogs on the Python side (both to save extra calls and to make final yaml or pickle files when done), it would make the most sense to me if the function returned a string key (hash or whatever) that could be saved on the Matlab side.

jeanconn:
I assume for proseco catalog creation we only need to handle the case of a single monitor window, but that should be defined.

jeanconn:
Do you want to handle errors with a full empty catalog or the like or handle errors in each of the catalog types such that we could return acqs without guides etc?

taldcroft:
Just wondering why obsid is not sufficient for a cache key.

jeanconn:
First, because I think the same obsid could be moved in the schedule during the same python session in matlab and have very different temperatures and maneuver angle. Second, because I don't think I have an obsid for "Create FOT Request" and I'm not entirely sure that I have obsids in general in the star selection data structure.

I thought it would be much safer to have the key based on the things that are actually used to determine the catalog.

taldcroft:
Ok.  Note that the key will only be an indicator that the cached catalog *might* be ok to use.  There will then be detailed checking that it *is* consistent with the new inputs to within tolerances.  But we can come up with some other key, the trick is letting it be slightly fuzzy.  Healpix pixel might be a start.

jeanconn:
> Note that the key will only be an indicator that the cached catalog might be ok to use. 

I don't quite understand this.  I figured we'd just set up caching to do that checking before putting catalogs into the cache.

taldcroft:
Not sure if we're on the same page.  Maybe this helps?
```
def func(**kwargs):
    key = get_key(kwargs)
    if key in cache and close_enough(cache[key], kwargs):
        out = cache[key]
        # Make some updates as needed, e.g. tweak yang/zang and row/col slightly
    else:
        # Do processing to generate `out`
        cache[key] = out
    return out
```
`close_enough()` inspects the `cache[key]` to find the kwargs that were used to generate it (in `meta`) and works from that.  To be conservative we could make `close_enough` be an exact match, if worried about the slight tweaking process.

taldcroft:
And that's a functional version, in practice would likely be class methods and a class-level cache.

jeanconn:
> Note that the key will only be an indicator that the cached catalog might be ok to use.

Still not sure what you mean.  If the tolerances are set where we want, the matching key will be an indicator that the catalog cached at that key is ok to use.  There is no "might be ok" in my thinking here.

taldcroft:
That might be a holdover from the concept of obsid as cache key.  But moving away from that you're likely correct.

jeanconn:
I'll also try to better understand the "don't ask for a new catalog if you already have one" logic that is already in the system.  There's a chance it might make sense to have the Matlab side provide something to use as a key instead of all of the parameters.

jeanconn:
I updated this to use an ACACatalogTable, but it looks like I haven't figured out pickling.  I removed the weakrefs (from the example in #59) for now as the weakrefs didn't pickle.  It looks like I can now pickle a dict of ACACatalogs, but that they don't unpickle.  I assume this is because my ACACatalog table has no rows or some such at this point, or maybe I've completely messed up the acqs/fids/guides meta assignment.


```

In [1]: from proseco.catalog import get_aca_catalog

In [2]: import pickle

In [3]: acas = {}

In [4]: acas[19387] = get_aca_catalog(19387)

In [6]: acas[18805] = get_aca_catalog(18805)

In [8]: mypickle = pickle.dumps(acas)

In [10]: pickle.loads(mypickle)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-768e29d6d271> in <module>()
----> 1 pickle.loads(mypickle)

/proj/sot/ska/jeanproj/git/proseco/proseco/core.py in __setstate__(self, state)
    190 
    191     def __setstate__(self, state):
--> 192         super().__setstate__(state)
    193         self.log_info = self.meta['log_info']
    194         del self.meta['log_info']

/fido.real/skare3/lib/python3.6/site-packages/astropy/table/table.py in __setstate__(self, state)
    419     def __setstate__(self, state):
    420         columns, meta = state
--> 421         self.__init__(columns, meta=meta)
    422 
    423     @property

/proj/sot/ska/jeanproj/git/proseco/proseco/core.py in __init__(self, data, print_log, **kwargs)
     69 
     70     def __init__(self, data=None, print_log=False, **kwargs):
---> 71         super().__init__(data=data, **kwargs)
     72         self.log_info = {}
     73         self.log_info['events'] = []

/fido.real/skare3/lib/python3.6/site-packages/astropy/table/table.py in __init__(self, data, masked, names, dtype, meta, copy, rows, copy_indices, **kwargs)
    406 
    407         # Finally do the real initialization
--> 408         init_func(data, names, dtype, n_cols, copy)
    409 
    410         # Whatever happens above, the masked property should be set to a boolean

/fido.real/skare3/lib/python3.6/site-packages/astropy/table/table.py in _init_from_dict(self, data, names, dtype, n_cols, copy)
    722 
    723         data_list = [data[name] for name in names]
--> 724         self._init_from_list(data_list, names, dtype, n_cols, copy)
    725 
    726     def _init_from_table(self, data, names, dtype, n_cols, copy):

/fido.real/skare3/lib/python3.6/site-packages/astropy/table/table.py in _init_from_list(self, data, names, dtype, n_cols, copy)
    687             cols.append(col)
    688 
--> 689         self._init_from_cols(cols)
    690 
    691     def _init_from_ndarray(self, data, names, dtype, n_cols, copy):

/fido.real/skare3/lib/python3.6/site-packages/astropy/table/table.py in _init_from_cols(self, cols)
    752         if len(lengths) != 1:
    753             raise ValueError('Inconsistent data column lengths: {0}'
--> 754                              .format(lengths))
    755 
    756         # Set the table masking

ValueError: Inconsistent data column lengths: set()
```

taldcroft:
I will look in to this, I have some experience here.

taldcroft:
The problem is that the `ACACatalogTables` don't have any *columns*.  So that is an upstream bug in Table that you can't pickle an entirely empty table.
```
In [18]: t = Table()

In [19]: pickle.loads(pickle.dumps(t))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
...
ValueError: Inconsistent data column lengths: set()
```
I would propose that the "merged" ACA catalog table has (to the extent possible) the same columns as a starcheck catalog from mica.starcheck.  But a minimal workaround for interface testing is just adding any column to the table before pickling, e.g.
```
acas[19387]['a'] = [0]
acas[18805]['a'] = [0]
```

proseco/catalog.py
```diff
@@ -0,0 +1,94 @@
...
+
+    merge_cat = merge_cats(fids=catalog.fids, guides=catalog.guides, acqs=catalog.acqs,
+                           mons=monitor_windows)
+    for col in merge_cat.colnames:
+        catalog[col] = merge_cat[col]
+
+    return catalog
+
+
+def merge_cats(fids=[], guides=[], acqs=[], mons=[]):
```
taldcroft: Should never use `[]` in a function def, this is a common gateway for difficult to understand bugs.  Use `None` instead.

proseco/catalog.py
```diff
@@ -0,0 +1,94 @@
...
+    # Start with just a list of entries and ids
+    entries = []
+    for fid in fids:
+        entries.append({'type': 'FID', 'id': fid['id']})
+    for bot_id in set(acqs['id']).intersection(guides['id']):
+        entries.append({'type': 'BOT', 'id': bot_id})
+    for gui_id in set(guides['id']) - set(acqs['id']):
+        entries.append({'type': 'GUI', 'id': gui_id})
+    for mon in mons:
+        entries.append({'type': 'MON', 'id': ''})
```
taldcroft: This will turn the column into `str` instead of `int`.  Maybe `-999`?  Not sure, but some number.

proseco/catalog.py
```diff
@@ -0,0 +1,94 @@
...
+    return catalog
+
+
+def merge_cats(fids=[], guides=[], acqs=[], mons=[]):
+
+    # Start with just a list of entries and ids
+    entries = []
+    for fid in fids:
+        entries.append({'type': 'FID', 'id': fid['id']})
+    for bot_id in set(acqs['id']).intersection(guides['id']):
```
taldcroft: Use `set(acqs['id']) & set(guides['id'])`

proseco/catalog.py
```diff
@@ -0,0 +1,94 @@
...
+                           mons=monitor_windows)
+    for col in merge_cat.colnames:
+        catalog[col] = merge_cat[col]
+
+    return catalog
+
+
+def merge_cats(fids=[], guides=[], acqs=[], mons=[]):
+
+    # Start with just a list of entries and ids
```
taldcroft: `.. types and ids`

taldcroft:
With regards to warnings and errors #75, how about the following for warnings for now in `ACACatalogTable`:
```
@property
def warnings(self):
    if 'warnings' not in self.meta:
        self.meta['warnings'] = []
    return self.meta['warnings']

def add_warning(self, warn):
    self.meta.warnings.append(warn)
```
Right now this just putting warnings on the individual tables.  The warnings can then be concatenated in the ACA table.   Note that `warn` can be a rich object with context, or whatever.

jeanconn:
Do the exceptions also need to be an attribute?  Or am I still confused and warnings will have exceptions?

taldcroft:
Originally in the traceback issue, but should be here.

> Regarding the empty tables, on the matlab side, I figured zero length would be fine to start (probably the thing we can most easily check anyway), so I just started with that in at ca55023

Re: the current stub (below) and the previous comment about a "zero length":
```
        if aca.guides is None:
            #aca.guides = GuideTable.empty()
            aca.guides = []
```
You had indicated the Matlab interface was roughly:
```
pyexec("guide_ids = aca.guides['id'], $guide_ids")  # I might be mis-remembering the exact syntax
```
The natural next step in Matlab is checking if the length of `guide_ids` is zero.

So it seems like you really want to have `aca.guides` always be a `GuideTable` with an `id` column.  I.e. have a consistent API where that object *always* presents the same return type (not `GuideTable` sometimes, or `list` other times).

taldcroft:
> Do the exceptions also need to be an attribute? Or am I still confused and warnings will have exceptions?

There can be only one exception since the only high-level try/except is in `get_aca_catalog`, which wraps the entire `_get_aca_catalog()` call.  It doesn't really need to be an attribute (property).

That is different from warnings, which are a list of warnings.

proseco/catalog.py
```diff
@@ -0,0 +1,146 @@
...
+    rows = []
+    slot = 0
+
+    for fid in fids:
+        fid['slot'] = slot
+        slot = (slot + 1) % 8
+        rows.append(fid[colnames])
+
+    for bot_id in set(acqs['id']) & set(guides['id']):
+        acq = acqs.get_id(bot_id)
```
taldcroft: Note the `get_id` method (and its friend `get_id_idx`).  These are quite helpful!

proseco/catalog.py
```diff
@@ -0,0 +1,146 @@
...
+    aca.fids = get_fid_catalog(detector=detector, sim_offset=sim_offset,
+                               focus_offset=focus_offset, acqs=aca.acqs,
+                               stars=aca.acqs.meta['stars'], print_log=print_log)
+
+    aca.guides = get_guide_catalog(obsid=obsid, att=att, date=date, t_ccd=t_ccd_guide,
+                                   n_guide=n_guide, dither=dither_guide,
+                                   stars=aca.acqs.meta['stars'], print_log=print_log)
+
+    # Make a merged starcheck-like catalog.  Catch any errors at this point to avoid
+    # impacting operational work (call from Matlab).
```
jeanconn: I still don't see a need for the merged table at all on the Matlab side, so maybe this should be a method on ACACatalogTable that we could call (optionally) if we want it on the SOT side.

jeanconn: PS My slot-determination code was a port of the matlab code and seemed to work.  I don't think this is right:
``

    0     1         1  FID  8x8   0.000    7.00    8.00   922.59 -1737.89     1     1    25
    1     2         5  FID  8x8   0.000    7.00    8.00 -1826.28   160.17     1     1    25
    2     3         6  FID  8x8   0.000    7.00    8.00   388.59   803.75     1     1    25
    3     4 157288312  BOT  6x6   0.739    8.73   10.23 -2143.88 -1717.81    20     1   160
    4     5 157689696  BOT  6x6   0.666    8.83   10.33  2445.48  -331.04    20     1   160
    5     6 157288520  GUI  6x6   0.000    8.81   10.31 -2394.12    45.51     1     1    25
    6     7 157692544  ACQ  6x6   0.685    9.02   10.52   148.06  1677.10    20     1   160
    7     8 157693576  ACQ  6x6   0.971    9.67   11.17  1744.25  1487.00    20     1   160
    0     9 157289008  ACQ  6x6   0.959    9.73   11.23   137.10 -2375.30    20     1    80
    1    10 157691632  ACQ  6x6   0.922    9.44   10.94  2296.85   167.54    20     1   160
    2    11 157288248  ACQ  6x6   0.979    9.61   11.11   -56.00 -1535.49    20     1   120
    3    12 157690680  ACQ  6x6   0.650    9.19   10.69  2444.36  1203.55    20     1   160
```

Which looks to me like slot assignment to a BOT and an ACQ in slot 3
Inputs
```
cat = proseco.catalog.get_aca_catalog(att=(20, 20, 30), man_angle=90, date='2018:001', dither_acq=(8,
    ...: 8), dither_guide=(8, 40), t_ccd_acq=-5, t_ccd_guide=-20, detector='ACIS-S', sim_offset=0, focus_offse
    ...: t=0, n_guide=3, n_fid=0, n_acq=8, print_log=True)
```

File: proseco/catalog.py [hunk 7]
```diff
@@ -0,0 +1,146 @@
...
+    for gui_id in set(guides['id']) - set(acqs['id']):
+        guide = guides.get_id(gui_id)
+        guide['slot'] = slot
+        slot = (slot + 1) % 8
+        rows.append(guide[colnames])
+
+    for acq_id in set(acqs['id']) - set(guides['id']):
+        acq = acqs.get_id(acq_id)
+        acq['slot'] = slot
+        slot = (slot + 1) % 8
```
jeanconn: I don't know how your brain works, but for me, the acq slots don't increment on guide stars so you need separate slot counters if you do it this way.  I think.

proseco/catalog.py
taldcroft: Sorry, you're right.  I went the other way to get the slot and type assigned back to the original tables (which I like), and couldn't immediately wrap my head around how do that with the original.  Will fix.

taldcroft: Agreed that Matlab doesn't need a merged table.  From the Matlab perspective the `aca` object is just a nice container that has the acqs, fids, and guides.  It doesn't care that there is a merged star catalog as well.

I do want the merged table to be in the pickle and to be handy for generating load products for validation testing and comparison with flight catalogs.  Do you see a problem with that?  I don't think there are really any downsides.

proseco/tests/test_catalog.py
```diff
@@ -0,0 +1,37 @@
...
+
+HAS_SC_ARCHIVE = Path(mica.starcheck.starcheck.FILES['data_root']).exists()
+
+
+@pytest.mark.skipif('not HAS_SC_ARCHIVE', reason='Test requires starcheck archive')
+def test_get_aca_catalog_19387():
+    """Put it all together.  Regression test for selected stars.
+    """
+    # Force not using a bright star so there is a GUI-only (not BOT) star
+    aca = get_aca_catalog(20603, exclude_ids=[40113544])
```
jeanconn: Nice use of exclude_ids.

proseco/catalog.py
taldcroft: Fixed the slot problem.  The new version is more complex but enforces consistency between the different tables (slots and entry ordering).  But of course I'm not sure if Matlab maintains the row ordering.  The goal here being that `aca.acqs[4]['slot']` actually matches the slots of the flight catalog for the 5th acquisition star.

proseco/tests/test_catalog.py
```diff
@@ -0,0 +1,37 @@
...
+import mica.starcheck
+
+from ..catalog import get_aca_catalog
+
+
+HAS_SC_ARCHIVE = Path(mica.starcheck.starcheck.FILES['data_root']).exists()
+
+
+@pytest.mark.skipif('not HAS_SC_ARCHIVE', reason='Test requires starcheck archive')
+def test_get_aca_catalog_19387():
```
jeanconn: Though I suppose the test name is a fib

proseco/catalog.py
taldcroft: Consistency in slot and type for a given AGASC ID, and a goal being to have the merged catalog be exactly the same as what we see in starcheck.

taldcroft:
Except for the top-level docstrings and function signature, I think this is ready for review.

Current profiling shows that get_aca_catalog takes 1200 msec, and the call to merge_cats is 25 msec.  So unless we get the rest of that way down, I would leave merge_cats in the default run path even though it is (mostly) wasted effort for Matlab.

jeanconn:
That's fine.  I mostly thought it was stealing focus from both of us in the 1.0 release.

proseco/catalog.py
jeanconn: OK, still not sure what that really means (with regard to changes in the code from my prototype).  The lack of option for monitors is probably the only thing that will keep these matching in starcheck anyway, but of course small number of observations.

jeanconn:
Also my [thumbs up PR](#79) is against this branch, so if you'd like to move along with it or something like it that would be good to know.

taldcroft:
> Another outstanding issue is that I'm not sure if the "halfwidths" for fid and guide should be passed from proseco or set from defaults on the matlab side.

Continue doing this from Matlab, assuming it is working with the current proseco that does not handle halfwidths for guide / fid.

jeanconn:
Was looking to simplify, but that's fine.

taldcroft:
@jeanconn - can you give a quick look?  Even if this isn't perfect it's probably time to merge to keep things manageable.

Note the last commit, which is the same as in your thumbs branch.  I think it won't conflict.

jeanconn:
I think this is fine.  May I merge the thumbs PR into this too and then add tests later (new PR)?

taldcroft:
Yes on thumbs merge

PR #66, opened 2018-09-08
Use mag instead of MAG_ACA consistently

taldcroft:
I think we should declare that the `mag` column of a `StarsTable` is the one that accurately reflects our best estimate of star magnitude.  Originally intended as a shorthand, this makes sense for future upgrades where we incorporate observational data to improve the star mags used in star selection.  This matches usage of `mag_err` as the best estimate of error.

Note: currently `mag_err` includes a term which is actually specific to acquisition, namely the per-readout standard deviation.  For guide selection that doesn't really make sense, and @malgosias has shown that the new catalog mag errors do match the observed distribution in guide star mean mags.

`MAG_ACA` would always reflect what is in the AGASC (mostly irrelevant in star selection), but would be there for reference as needed.

### DIFF
--- a/proseco/acq.py
+++ b/proseco/acq.py
@@ -184,12 +184,12 @@ class AcqTable(ACACatalogTable):
         :returns: Table of candidates, indices of rejected stars
         """
         ok = ((stars['CLASS'] == 0) &
-              (stars['MAG_ACA'] > 5.9) &
-              (stars['MAG_ACA'] < 11.0) &
+              (stars['mag'] > 5.9) &
+              (stars['mag'] < 11.0) &
               (~np.isclose(stars['COLOR1'], 0.7)) &
               (np.abs(stars['row']) < CHAR.max_ccd_row) &  # Max usable row
               (np.abs(stars['col']) < CHAR.max_ccd_col) &  # Max usable col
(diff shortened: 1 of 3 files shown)


jeanconn:
Regarding "Note: currently mag_err includes a term which is actually specific to acquisition, namely the per-readout standard deviation.", I didn't think the per-readout standard deviation was specific to acquisition.  I thought this per-readout std was somewhat relevant for the guide filter trying to get stars that are going to have read-out-magnitudes within the specified mag range.  Probably less useful when trying to compare a star's mag to its possible spoiler.

taldcroft:
Indeed one needs to look at the details of tests where `mag_err` is being used to assess which version to use.  I had started with the case you mentioned of nearby spoilers, where the mean mag err is more relevant, but haven't thought about other selection criteria.  In any case, this is somewhat of a 2nd-order detail, so I'm opening a low-priority issue for later.

jeanconn:
Do you plan something similar with "color"?

taldcroft:
Good question.  Since we now have up to 3 color values for a star, maybe it is best to keep those explicit.  But I haven't thought about it carefully.

taldcroft:
See #70 to re-close this issue.

Issue #50, opened 2018-08-29
What's the recommended path to get acq yaml/reporting on manual catalog?


jeanconn:

taldcroft:
Would it be feasible to have Matlab call an `AcqTable` method to do a manual edit, or more likely update the catalog after a series of edits which are confirmed?

This gets to idea that has been percolating that the backend Python session is persistent, so the `AcqTable` objects for all the obsids in the schedule can likewise be persistent and can be manipulated thusly.

So if there is just a method saying effectively "here is the new catalog after manual edits", then that method can handle the implementation details of ensuring that the YAML is consistent.

taldcroft:
This persistence is probably a practical requirement if you want to allow things like a button to tell it to create the report for a particular obsid.

jeanconn:
I thought when we were talking about running the calculations on the non-optimized catalog that there was already a method to just pass the needed pieces back to get this calculation.  Wasn't sure if I was just missing it.

I think I can pretty easily determine if the catalog is a 'saved' catalog or an automatically generated catalog, and I thought that 'saved'/manual might be enough (not if the the acq catalogs were changed, just if the final catalog had been saved after editing with any kind of change).

I wasn't planning to use persistent data in the Python session, but could work from that angle if need be.  In my first go, I've just saved the yaml text into the obsid starResults object.

jeanconn:
And I'm still editing with the prospect of not making any buttons.

taldcroft:
Generating the YAML is quite slow so it should be done only if required.  (It's slow because of all the flattening of table and numpy objects into pure Python).

taldcroft:
*Eventually* there will be buttons.  BTW, on my old slow iMac:
```
In [3]: time acqs = acq.AcqTable.get_acq_catalog(19387)
CPU times: user 1.24 s, sys: 33.7 ms, total: 1.27 s
Wall time: 1.33 s

In [4]: time out = acqs.to_yaml()
CPU times: user 2.41 s, sys: 28.2 ms, total: 2.44 s
Wall time: 2.44 s
```

jeanconn:
OK, so we will need to work more on the best ways to optimize.  

In my thinking, if they run star selection once with YAML saving and then we cache it on the Matlab side so the catalogs and YAML don't change unless the attitudes change (which I think is basically already done), there won't be much rework or additional slowness. 

I didn't see any way to get the YAML without making the YAML.

taldcroft:
[...]

Issue #26, opened 2018-08-16
Return max ACA temperature for catalog


taldcroft:
This can be used within ORviewer / MCC for dynamic ACA limits in schedule planning.  This should use the "classic" acquisition planning limit of "Probability of 2 or fewer stars" using the marginalized p_acq values for the acq catalog stars.

Operational planning use of this feature needs to include a substantial increase of the ACA planning limit, perhaps 2 C.  This allows new planning real estate, but without immediately allowing very high ACA temperatures.  Then as Sabina mentioned, the dynamic limit looks more like chopping out chunks of a higher limit (for bad catalogs) than adding chunks to a lower limit (for good catalogs).

PR #69, opened 2018-09-09
Include and exclude stars from catalog selection

### LINKED ISSUE #3: Starcheck needs to be able to call to get access to probabilities
There needs to be a pathway for starcheck to feed in a load catalog (which may have been hand-edited) and get the individual star marginalized `p_acq` and probably `p_safe`.  Right now the code assumes the acq catalog has been generated and has the required metadata.

taldcroft:
This allows for including and excluding stars from catalog selection.

The immediate use is for starcheck, where it could be used to force the entire catalog to be what is 
found in backstop.  @jeanconn - would this meet your requirements?  It will regenerate probabilities so that the `p_acq` needed in starcheck (the marginalized versions going forward) will be available as `acqs['p_acq']` after re-getting the catalog with 8 appropriate include stars set.  (I realize now we need a param for max acq stars to force a catalog with fewer than 8).

Closes #3 
Addresses #50

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -113,8 +113,11 @@ class ACACatalogTable(Table):
             assert self['id'][idx] == id
         except (KeyError, IndexError, AssertionError):
             self.make_index()
-            idx = self._id_index[id]
-            assert self['id'][idx] == id
+            try:
+                idx = self._id_index[id]
+                assert self['id'][idx] == id
+            except (KeyError, IndexError, AssertionError):
+                raise KeyError(f'{id} is not in table')
 
         return idx
 
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -1,3 +1,4 @@
+import warnings
 import pickle
 import inspect
 import time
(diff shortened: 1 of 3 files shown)


taldcroft:
BTW this highlights use of `StarsTable` fake stars capability to just make exactly the stars you want and test in a well-defined and faster way (no getting stars from AGASC and working with much smaller numbers of stars in the catalog).

proseco/core.py
```diff
@@ -113,8 +113,11 @@ def get_id_idx(self, id):
             assert self['id'][idx] == id
         except (KeyError, IndexError, AssertionError):
             self.make_index()
-            idx = self._id_index[id]
-            assert self['id'][idx] == id
+            try:
```
taldcroft: Somewhat unrelated but needed this.

jeanconn:
I think this will work.  Though a problem I'm seeing is that I *think* if you also want to do #26 that I can't directly use t_ccd_warm_limit:

https://github.com/sot/chandra_aca/blob/master/chandra_aca/star_probs.py#L71

which wants to just calculate probabilites (and temperature from that) based on color/halfwidth/mag etc.  So I assume we'll also need something else for that?

taldcroft:
Good catch.  The warm limit calculation will likely need to be a method of AcqTable. Implementation TBD.

PR #79, opened 2018-09-13
Add actual thumbs_ups

jeanconn:
Just an idea for some actual thumbs-up checking.

### DIFF
--- a/proseco/catalog.py
+++ b/proseco/catalog.py
@@ -89,6 +89,9 @@ def _get_aca_catalog(**kwargs):
     aca.fids = get_fid_catalog(acqs=aca.acqs)  # FidTable grabs relevant kwargs from ``acqs.meta``
     aca.guides = get_guide_catalog(stars=aca.acqs.meta['stars'], **guide_kwargs)
 
+    # Get overall catalog thumbs_up
+    aca.thumbs_up = aca.acqs.thumbs_up & aca.fids.thumbs_up & aca.guides.thumbs_up
+
     # Make a merged starcheck-like catalog.  Catch any errors at this point to avoid
     # impacting operational work (call from Matlab).
     try:
--- a/proseco/characteristics.py
+++ b/proseco/characteristics.py
@@ -97,3 +97,7 @@ bad_star_list = [36178592,
                  1196953168,
                  1197635184]
 
+# Minimum acquisition probability thresholds from starcheck
+# for thumbs_up
+acq_prob_n = 2
+acq_prob = 8e-3
--- a/proseco/characteristics.py
+++ b/proseco/characteristics.py
@@ -97,7 +97,6 @@ bad_star_list = [36178592,
                  1196953168,
                  1197635184]
 
-# Minimum acquisition probability thresholds from starcheck
-# for thumbs_up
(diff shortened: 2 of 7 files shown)


proseco/core.py
```diff
@@ -82,6 +82,7 @@ def __init__(self, data=None, print_log=False, **kwargs):
             self._default_formats[name] = '.2f'
         for name in ('ra', 'dec', 'RA_PMCORR', 'DEC_PMCORR'):
             self._default_formats[name] = '.6f'
+        self.thumbs_up = False
```
jeanconn: Looks l don't really have booleans on the matlab side and the python interface won't pass a numpy.bool type so I end up converting to int anyway.  Should probably just store these as ints on the python side.

proseco/catalog.py
```diff
@@ -54,11 +54,16 @@ def _get_aca_catalog(obsid=0, att=None, man_angle=None, date=None,
     aca.fids = get_fid_catalog(detector=detector, sim_offset=sim_offset,
                                focus_offset=focus_offset, acqs=aca.acqs,
                                stars=aca.acqs.meta['stars'], print_log=print_log)
+    # Set the fid thumbs up to true for now as I'm not sure what to check
```
taldcroft: Thumbs up for fid can be if it `len(fids) == fids.meta['n_fid']`.  Set it in `get_fid_catalog`.

proseco/core.py
taldcroft: As always, leave a vertical line of space between entirely unrelated blocks (or lines) of code.

taldcroft: Add a `thumbs_up` property which sets onto the same-named meta key.
```
@thumbs_up.setter
def thumbs_up(self, val):
    self.meta['thumbs_up'] = int(val)  # Coerce to int for interface with Matlab
```

proseco/characteristics.py
```diff
@@ -97,3 +97,7 @@
                  1196953168,
                  1197635184]
 
+# Minimum acquisition probability thresholds from starcheck
+# for thumbs_up
```
taldcroft: Doesn't need to be split into two lines.

taldcroft:
Oops, I forgot to add that it needs tests.

proseco/core.py
jeanconn: I thought thumbs_up already was a property set to the same-named meta key?  Wouldn't this just be a change to for the type?

taldcroft: Indeed...

taldcroft:
Not sure if you deleted that previous comment, but yes there is an `n_fid` in the PR (though it is mistakenly called `n_fids`).  But for this PR maybe just stub the fid thumbs up to True (i.e. ignore my previous suggestion)?

jeanconn:
Which PR has n_fid or n_fids?  I'm not seeing where this is handled.  (fine to go back to just setting it thumbs_up true for fid catalog but trying to be situationally aware on progress and interface).

jeanconn:
Gotcha. Yes, for integration I wasn't sure if get_aca_catalog needed to pass n_fid to get_acq... to then have it get in acq meta etc... but I think that doesn't drive any changes to my top level API (as long as I can assume that we only have the requested fids by the time it gets to the matlab interface).

PR #80, opened 2018-09-13
Make one guide test faster

taldcroft:
@jeanconn - can we do this?  That one test is pretty slow.  I'm guessing that test coverage is not reduced by this change, but not sure.

### DIFF
--- a/proseco/tests/test_guide.py
+++ b/proseco/tests/test_guide.py
@@ -211,8 +211,8 @@ def test_check_mag_spoilers():
 
     # The mag spoiler check only works on stars that are within 10 pixels in row
     # or column, so don't bother simulating stars outside that distance
-    r_dists = np.arange(-9.25, 10, 1)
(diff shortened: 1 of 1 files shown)


proseco/tests/test_guide.py
```diff
@@ -211,8 +211,8 @@ def test_check_mag_spoilers():
 
     # The mag spoiler check only works on stars that are within 10 pixels in row
     # or column, so don't bother simulating stars outside that distance
-    r_dists = np.arange(-9.25, 10, 1)
-    c_dists = np.arange(-9.5, 10, 1)
+    r_dists = np.arange(-9.25, 10, 2)
+    c_dists = np.arange(-9.5, 10, 2)
```
jeanconn: That's fine.  The bigger the steps, the less sensitive the test, but if we modify the algorithm we can always decrease the step size again for validation.

PR #81, opened 2018-09-13
 Support specifying number of fids

taldcroft:
This has partial support for specifying the number of fids by simply chopping down the initial catalog list.  But I think it supports interface and test requirements for 1.0. 

This also adds tests of the `thumbs_up` attribute.


proseco/fid.py
```diff
@@ -18,7 +18,7 @@
 
 
 def get_fid_catalog(*, detector=None, focus_offset=0, sim_offset=0,
-                    acqs=None, stars=None, dither=None,
+                    acqs=None, stars=None, dither=None, n_fid=3,
```
jeanconn: Somewhat in the weeds for interface, but do you want to be able to pass this n_fids with the acqs meta?  I suppose the block that assigns parameters from acqs.meta could conditionally override if n_fid != 3.  And you'd need to add n_fid to the acq params in get_aca_catalog.

taldcroft: I thought about that and decided against that for now.  But it doesn't matter for Matlab and we can revisit.

Issue #89, opened 2018-09-16
No fids result in empty fid table without any columns (no 'id' column either)


jeanconn:

jeanconn:
Do we need something similar for get_guide_catalog?  I didn't understand if I was supposed to infer from your text in #92 @taldcroft  that, because get_guide_catalog will probably throw an exception before it will return an empty table, we don't need to do anything.

taldcroft:
For obvious things we can think of, guide should return an empty table and not throw an exception.  That way you get the log and anything else along the way.  Top level exception catching short really be a last resort.

PR #90, opened 2018-09-16
Replace acq/gui/fid with empty table if needed

jeanconn:
Replace acq/gui/fid with empty table if needed

This is a possible solution for #89


jeanconn:
I'm assuming there aren't any "empty table" tests (or would have notice this before) so this probably could use a test.

jeanconn:
Superceded by #94 #92 #93

Issue #55, opened 2018-09-05
Fid selection should watch out for spoiler stars in the fid box with (guide phase) dither


jeanconn:

taldcroft:
I think this is done, but am I misinterpreting??  Note that the code in master at this point does not include *using* the spoiler score to co-optimize fid and acq catalogs.  See [page 21 - 29](https://docs.google.com/presentation/d/1j89Zy2RRB4McV32ue-EpQThaEpiONw_EAilqcOqz45g/edit#slide=id.gc8be552d7_0_0)

https://github.com/sot/proseco/blob/977466d787a3e25b4b0e24741191f469d88d659c/proseco/fid.py#L302

taldcroft:
OK, I put in #56 now.  Can this issue be closed?

PR #91, opened 2018-09-16
Add ACABox class and use 2-component dither in acq, fid selection

### LINKED ISSUE #56: Use dither_y and dither_z by acq phase and guide phase throughout
Scrub all acq, guide, and fid selection to ensure proper use of two-component dither.

taldcroft:
This adds a class `ACABox` that can be used to hold a dither.  Depending on how the code goes I think it might hold something that isn't strictly a dither, so it has a generic name.

Status:
- [x] Pass guide tests
- [x] Pass acq and fid tests using a stub `max()` method that gives the same scalar currently used in the code
- [x] Fix acq and fid code to use actual `y` and `z` component values
- [x] Write new tests (e.g. large dither obs) with different `y` and `z` values.

Closes #56

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -96,6 +96,10 @@ class ACABox:
         return not self == other
 
     def __gt__(self, other):
+        """
+        Returns True if either component is larger.  This is somewhat
+        specific to how boxes are used in intruder processing.
+        """
         if not isinstance(other, ACABox):
             other = ACABox(other)
         return self.y > other.y or self.z > other.z
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -53,7 +53,7 @@ class ACABox:
         either a two-element sequence (y, z) or a scalar (which applies
         to both y and z.
 
-        :param dither: scalar or 2-element sequence (y, z), (arcsec, default=20)
+        :param size: scalar or 2-element sequence (y, z), (arcsec, default=20)
 
         """
         if isinstance(size, ACABox):
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -46,6 +46,53 @@ def to_python(val):
     return val
 
 
+class ACABox:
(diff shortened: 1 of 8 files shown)


proseco/core.py
```diff
@@ -46,6 +46,53 @@ def to_python(val):
...
+        except AssertionError:
+            # Has a length but it is not 2
+            raise ValueError('size arg must be either a scalar or a two-element sequence')
+        else:
+            self.y = size[0]
+            self.z = size[1]
+
+    @property
+    def row(self):
+        return self.y / 5
```
taldcroft: This does still use 5.0 as the pixel scale.  As discussed previously this seems fine to me and makes things tidy for acq where I want 20 pixels to correspond to precisely 100 arcsec for all the fixed-grid values like search box size (60, 80, 100, ...).  Using 4.97... could be made to work too (maybe even right now), but not sure what it buys us.  Even the more exact value in the guide characteristics is an approximation since the plate scale varies with position.

taldcroft:
For the record, here is a report image with 8x64 dither.  Looks right to me.
image

jeanconn:
Though just once more for the record, the acq boxes will never be 8x64 in our current operational strategy.  I'm trying to recall if there are tests for the fids that would show that the 8x64 dither is used when looking for the intruder but not the imposter; will review.

jeanconn:
Yes, though I tried to make the point that fid selection needs to check for spoiler stars with guide phase dither in #55 and you wanted that closed , to be handled in  #56 which you've now closed even though it just handled half of the issue (using a box not handling both phases for fids... if I'm understanding where the code is at this point).  So I don't know if this needs a new issue or to reopen a closed one.

taldcroft:
Thanks for your persistence.  I think this is finally handled in #95.

PR #94, opened 2018-09-17
Handle empty candidate list for guide and test for get_aca_catalog

jeanconn:
Handle empty candidate list for guide and add empty cand test for get_aca_catalog

### DIFF
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -134,6 +134,9 @@ class GuideTable(ACACatalogTable):
         """
         cand_guides = self.meta['cand_guides']
         self.log("Starting search stages")
+        if len(cand_guides) == 0:
+            self.log("There are no candidates to check in stages.  Exiting")
+            return cand_guides
         cand_guides['stage'] = -1
         n_guide = self.meta['n_guide']
         for idx, stage in enumerate(GUIDE_CHAR.stages, 1):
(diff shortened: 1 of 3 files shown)


proseco/guide.py
```diff
@@ -134,6 +134,9 @@ def run_search_stages(self):
         """
         cand_guides = self.meta['cand_guides']
         self.log("Starting search stages")
+        if len(cand_guides) == 0:
```
jeanconn: @taldcroft I looked at adding similar tweaks to what you used in proseco.acq, but for the empty candidate list, it seemed just as reasonable to stop running the checks and return the empty table of candidates.  Does this make sense?

As far as I could tell, not *selecting* any stars later through the search stages resulted in just the all-False mask within cand_guides, which returned a 0-list table with an 'id' column defined... which seemed all good.

taldcroft:
With this can we close #90?

taldcroft:
And maybe cut 1.0.2 after this?  I'm going to say this is a bug fix.

jeanconn:
Yes, I have a couple of edits here but then this + #92 + #93 supercede #90 and will make 1.0.2.

proseco/guide.py
```diff
@@ -134,6 +134,9 @@ def run_search_stages(self):
...
+            self.log("There are no candidates to check in stages.  Exiting")
+            return cand_guides
```
jeanconn: One edit.  I just should have a better comment here before return the empty cand_guides as the empty 'selected' set.  Could also bump this if statement back up one level and have a conditional on bothering to get here.

PR #95, opened 2018-09-18
Unify code, add MetaAttribute descriptor, and fix fid select to use relevant dither

taldcroft:
It all started with the realization that fid selection needs to use `dither_guide` to check for stars that spoil a fid, and `dither_acq` to check if a fid spoils and acq.  Implementing this would have made a somewhat ugly situation even worse.  

So against my better judgement I got pulled down the slippery slope of just making all the fixes to unify the acq, guide, and fid selection parameters via a new `MetaAttribute` descriptor, and an hour or two dragged into a very long morning to get everything working again.  Overall I think the code is a lot cleaner now, especially in `catalog.py` which was mostly deletions.

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -247,6 +247,22 @@ class ACACatalogTable(Table):
             if not isinstance(dither, ACABox):
                 setattr(self, dither_attr, ACABox(dither))
 
+    def set_stars(self, acqs=None):
+        """Set the object ``stars`` attribute to an appropriate StarsTable object.
+
+        If ``acqs`` is defined that will be a previously computed AcqTable with
+        ``stars`` already available, so use that.
+
+        :param acqs: AcqTable for same observation
+        """
+        if acqs is None:
+            if self.stars is None:
+                self.stars = StarsTable.from_agasc(self.att, date=self.date, logger=self.log)
+            else:
+                self.stars = StarsTable.from_stars(self.att, self.stars, logger=self.log)
+        else:
+            self.stars = acqs.stars
+
     @property
     def dither(self):
         super().__getattr__(self, 'dither')
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -159,6 +159,8 @@ class ACACatalogTable(Table):
     # Catalog attributes, gets set in MetaAttribute
     allowed_kwargs = set(['dither', 't_ccd'])
 
+    required_attrs = ('dither_acq', 'dither_guide', 'date')
+
     obsid = MetaAttribute(default=0)
     att = MetaAttribute()
     n_acq = MetaAttribute(default=8)
@@ -243,6 +245,13 @@ class ACACatalogTable(Table):
                     dither_z_amp = obso.get('dither_z_amp')
                     if dither_y_amp is not None and dither_z_amp is not None:
                         setattr(self, dither_attr, ACABox((dither_y_amp, dither_z_amp)))
+
+                        # Special rule for handling big dither from mica.starcheck,
+                        # which does not yet know about dither_acq vs. dither_guide.
+                        # Use the most common dither size for ACIS / HRC.
+                        if dither_attr == 'dither_acq' and self.dither_acq.max() > 30:
+                            dither = 8 if self.detector.startswith('ACIS') else 20
+                            self.dither_acq = ACABox(dither)
 
         for dither_attr in ('dither_acq', 'dither_guide'):
             dither = getattr(self, dither_attr)
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -199,7 +199,7 @@ class ACACatalogTable(Table):
         for name in ('ra', 'dec', 'RA_PMCORR', 'DEC_PMCORR'):
             self._default_formats[name] = '.6f'
 
-    def set_kwargs(self, **kwargs):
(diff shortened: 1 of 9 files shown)


proseco/core.py
```diff
@@ -156,6 +197,96 @@ def __init__(self, data=None, print_log=False, **kwargs):
...
+            if self.n_guide is None:
+                fid_or_mon = (obs['cat']['type'] == 'FID') | (obs['cat']['type'] == 'MON')
+                self.n_guide = 8 - np.count_nonzero(fid_or_mon)
+
+            for dither_attr in ('dither_acq', 'dither_guide'):
+                if getattr(self, dither_attr) is None:
+                    dither_y_amp = obso.get('dither_y_amp')
+                    dither_z_amp = obso.get('dither_z_amp')
+                    if dither_y_amp is not None and dither_z_amp is not None:
+                        setattr(self, dither_attr, ACABox((dither_y_amp, dither_z_amp)))
```
jeanconn: Somewhat in the weeds for this PR, but you could us the rules from the matlab tools to set this up to run on big dither observations as well to have working tests from mica.starcheck catalogs for those.   So, if either dither axis > 30?, set dither_acq to be square dither appropriate for the detector (20,20 for hrc etc).

proseco/core.py
```diff
@@ -156,6 +197,96 @@ def __init__(self, data=None, print_log=False, **kwargs):
...
+
+            obs = get_starcheck_catalog(self.obsid)
+            obso = obs['obs']
+
+            if self.att is None:
+                self.att = [obso['point_ra'], obso['point_dec'], obso['point_roll']]
+            if self.date is None:
+                self.date = obso['mp_starcat_time']
+            if self.t_ccd is None:
+                self.t_ccd = obs.get('pred_temp', -15.0)
```
jeanconn: How does this now get assigned for acq and guide?  Or does it not need to be set separately in this paradigm?

jeanconn: Wait.  Found it below so just deleting both comments.

proseco/core.py
```diff
@@ -156,6 +197,96 @@ def __init__(self, data=None, print_log=False, **kwargs):
...
+
+    def set_stars(self, acqs=None):
+        """Set the object ``stars`` attribute to an appropriate StarsTable object.
+
+        If ``acqs`` is defined that will be a previously computed AcqTable with
+        ``stars`` already available, so use that.
+
+        :param acqs: AcqTable for same observation
+        """
+        if acqs is None:
```
jeanconn: A little confused by API here.  If I want to pass stars to get_guide_catalog now, do I need to pass them via acqs?

jeanconn: OK.  I think I get it, I *could* pass them via acqs, but when acqs is none, then you check to see if stars is defined below and continue.  OK that makes sense.  And then in your standard interface for get_aca_catalog, you're just defining stars=aca.acqs.stars which is fine (and matches the docstring for get_guide_catalog).

proseco/guide.py
```diff
@@ -38,66 +37,20 @@ def get_guide_catalog(obsid=0, att=None, date=None, t_ccd=None, dither=None, n_g
...
-    if not isinstance(dither, ACABox):
-        dither = ACABox(dither)
-
-    if dark is None:
+    if guides.dark is None:
         from mica.archive.aca_dark import get_dark_cal_image
-        guides.log(f'getting dark cal image at date={date} t_ccd={t_ccd:.1f}')
-        dark = get_dark_cal_image(date=date, select='before', t_ccd_ref=t_ccd, aca_image=True)
+        guides.log(f'getting dark cal image at date={guides.date} t_ccd={guides.t_ccd:.1f}')
+        guides.dark = get_dark_cal_image(date=guides.date, select='before',
```
jeanconn: Doesn't matter at all for operational use but we've got a 'before' here and a 'nearest' in acq.

proseco/fid.py
```diff
@@ -36,19 +34,19 @@ def get_fid_catalog(*, detector=None, focus_offset=0, sim_offset=0,
     :param focus_offset: SIM focus offset [steps] (default=0)
     :param sim_offset: SIM translation offset from nominal [steps] (default=0)
     :param acqs: AcqTable catalog.  Optional but needed for actual fid selection.
-    :param stars: stars table.  Defaults to acqs.meta['stars'] if available.
+    :param stars: stars table.  Defaults to acqs.stars if available.
     :param dither: dither (float or 2-element sequence (dither_y, dither_z), [arcsec]
-                   Defaults to acqs.meta['dither'] if available.
+                   Defaults to acqs.dither if available.
```
jeanconn: So should this have both dithers in the docstring?

proseco/core.py
jeanconn: And I just checked the 30 number in characteristics and it matches.

The code I was mentioning that is going to do this from the Matlab side is 

https://github.com/sot/fot_matlab_tools/blob/master/MissionScheduling/schedule/ORs/UpdateORStarSelectors.m#L73

but that part of characteristics is not in our repo.

proseco/guide.py
taldcroft: Fixed.  (It was in report.py not acq.py though).

proseco/fid.py
taldcroft: Fixed.

PR #98, opened 2018-09-20
Make from starcheck text work for ERs and supply defaults

taldcroft:


proseco/core.py
```diff
@@ -1001,7 +1001,14 @@ def get_kwargs_from_starcheck_text(obs_text):
     # Remove common leading whitespace
     obs_text = textwrap.dedent(obs_text)
 
-    kw = {}
+    # Set up defaults
+    kw = {'dither': (8.0, 8.0),
```
jeanconn: Though for ERs, if this is for review of already-done observations, I think we want the state-based last-commanded dither.  And we should just add the dither to starcheck output.

PR #108, opened 2018-09-24
Remove property setting for print_log MetaAttribute

jeanconn:
Remove property setting for print_log MetaAttribute


jeanconn:
Not sure I completely understand the MetaAttributes, but removing the code that handled it as a property seems to make the print_log work again.

PR #109, opened 2018-09-24
Properly handle fid sets with fewer than 3 fids

### LINKED ISSUE #107: Handle 2-fid and 1-fid case
The 2-fid case can arise in (at least) a couple of ways:
- `n_fid=2` requested (e.g. Venus)
- `n_fid=3` requested but only 2 fids are on the CCD (ACIS undercover)

Any others?  The code needs to handle this gracefully.  Currently I don't believe either of these cases are handled adequately.

I don't see any obvious real-life use case for the 1-fid case, but it is easy enough once 2-fid is handled.

taldcroft:
Closes #107 

To do:
- [x] More tests
- [x] Fix `optimize_acqs_fids()` to use cand_fid_sets


taldcroft:
@jeanconn - you might want to look at 487db45.  Would have been good to break this out to a separate PR but didn't want to slow down.

taldcroft:
Looking myself at that thumbs-up change, I added 12d8822 to make them all symmetric.  The idea is that if user specifically requests 0 fids, or guides, or acqs, then getting a zero-length result is OK.  In practice this only applies to tests since in an operational catalog we would never have `n_guide=0`.

jeanconn:
Regarding the dynamic thumbs_up, were there "gotchas" in the previous code/setup?  I hadn't thought of a case where you could end up with a thumbs_up value that did not represent your catalog (unless manually set to the wrong value).

taldcroft:
I hit it for acq and fid, where the catalogs are updated and the code did not update thumbs_up accordingly.  So instead of whacking that mole in one place I decided the only safe thing is to compute dynamically.

Just thinking a little, I can see the case for keeping the thumbs_up property really being a well-defined operational thing, and failing for a catalog even if it is called with `n_guide=0`.  I had made the change so my tests would pass...

jeanconn:
> I hit it for acq and fid, where the catalogs are updated and the code did not update thumbs_up accordingly. 

I'm just trying to make sure I understand how the acq catalog (for example) was updated outside of "get_acq_catalog", which I thought set the value of thumbs_up at the end of the method.

taldcroft:
Acqs and fids are being updated in `proseco.catalog.optimize_acqs_fids()`.  Indeed `get_acq_catalog()` set thumbs-up at the end, and the value there was stale after `optimize_acqs_fids()`.  More obvious (and what got caught in testing) was that the fids thumbs-up was initially `0` because there was no "perfect" catalog initially.

Conceptually, the `catalog` module (`get_aca_catalog`) is doing the necessary integration work and `get_acq_catalog` and `get_fid_catalog` are just the first pass.

PR #117, opened 2018-09-26
Fix minor regression from #105

taldcroft:
Sorry, didn't notice this minor change in star order introduced by #105.


jeanconn:
Does this mean I forgot to run a test or was this a merge conflict on the test?

taldcroft:
I didn't track down why this happened, but there were no merge conflicts per se.

jeanconn:
By "merge conflict" in this case, I meant that maybe your test was added after my code change, so the test didn't work when merged (but git wouldn't have complained).

Somewhat relatedly I'm getting 6 test failures in master at b8c6125  .  I'm guessing some AGASC 1.7 issues?

jeanconn:
Yes, I set get_agasc_cone in my local proseco.core to use the 1.6 miniagasc, and the failures went away.  What's your proseco vs agasc plan?  Does your laptop need update and then we'll just "say" we're using 1.7?  Or do you want to control this in proseco?

taldcroft:
See 5ab32a5d.  That will likely cure the 1.6 vs 1.7 problem, assuming the small mag diffs don't change selection (though it may well do so).  I decided that star / fid id (and halfw, etc) is sufficient for regression tests of catalogs.

I will soon update my laptop and other platforms and just "say" we're using 1.7.

taldcroft:
Might be worth you cherry picking 5ab32a5d and seeing what happens in master with 1.7.  I just did it in that PR for time.

jeanconn:
Though there can still be selection order changes and such in guide selection.  I don't know if we should have a test that checks the miniagasc version in `agasc` or in `proseco`.  Could also explicitly make proseco tests work with either miniagasc (which is an effect of 5ab32a5 ).

jeanconn:
5ab32a5 doesn't appear to fix the test fails (3 acq, 2 catalog, 1 guide).

taldcroft:
Rats.  On the plus side, that means AGASC 1.7 is having an impact!

jeanconn:
I'm not sure "breaks our tests" is the impact we are shooting for, but... :smile:

PR #122, opened 2018-09-26
Change regression test outputs to be less fragile and work with AGASC 1.7

taldcroft:
The original regression test outputs would break for inconsequential changes.


jeanconn:
For me, I've got the issue that my "big dither" test has a star in the field that has changed a lot from AGASC 1.6 to 1.7

```
In [8]: agasc.get_star(977930352, agasc_file='/proj/sot/ska/data/agasc/miniagasc_1p7.h5')['MAG_ACA']
Out[8]: 8.6076403

In [9]: agasc.get_star(977930352, agasc_file='/proj/sot/ska/data/agasc/miniagasc_1p6.h5')['MAG_ACA']
Out[9]: 10.925166
```

Do you have a recommendation for what to do with the test?  For the test should I just fetch stars and manually set the value of MAG_ACA and MAG_ACA_ERR to be the 1.7 values?

taldcroft:
Sorry, you lost me.  All tests are passing with 1.7 now.

jeanconn:
OK.  I had a different idea of fragile.  I thought you wanted the tests to pass with agasc 1.6 or 1.7 (which wouldn't work for the big dither test).

taldcroft:
I suspect it is not possible to get all tests to be 1.6 and 1.7 compatible.  The acq star selection code is just too twitchy.  So proseco officially requires 1.7 to pass tests, though it will still run perfectly well and give valid outputs with 1.6.

PR #126, opened 2018-09-26
Fix a small test fail

taldcroft:


taldcroft:
I confirm that master @ 109b989 is passing tests now.  Will tag 1.1.

PR #128, opened 2018-09-27
Cut box sizes corresponding to max non-zero man_err

taldcroft:
This means that if there is zero probability of a man_err > 60" (for example), then no box size will be larger than 60".  This only affects the 0-5 and 5-20 deg man angle bins.  A follow-up PR will set the 0-5 deg man_err to be 0.0 for man_err > 60.


taldcroft:
Note that the only affect catalogs are those with small maneuver angles.

PR #134, opened 2018-09-28
Add more voluminous debug information for why each star was rejected in guide stages

jeanconn:
Add more voluminous debug information for why each star was rejected in guide stages

Also updates box spoiler test with star actually in box in 851b11e
And reorganizes determination of brightest 2x2 to aid recording where it was.

### DIFF
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -84,6 +84,14 @@ class GuideTable(ACACatalogTable):
 
     dark = MetaAttribute()
     cand_guides = MetaAttribute(is_kwarg=False)
+    reject_info = MetaAttribute(default=[], is_kwarg=False)
+
+    def reject(self, reject):
+        """
+        Add a reject dict to self.reject_info
+        """
+        reject_info = self.reject_info
+        reject_info.append(reject)
 
     @property
     def t_ccd(self):
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -235,7 +235,10 @@ class GuideTable(ACACatalogTable):
             ok = ok & ~bg_spoil & ~reg_spoil
 
         # Check for column spoiler
-        col_spoil = check_column_spoilers(cand_guides, ok, stars, n_sigma)
+        col_spoil, col_rej = check_column_spoilers(cand_guides, ok, stars, n_sigma)
+        for rej in col_rej:
+            rej['stage'] = stage['Stage']
+            self.reject(rej)
         cand_guides[scol][col_spoil] += GUIDE_CHAR.errs['col spoiler']
         ok = ok & ~col_spoil
 
@@ -455,6 +458,7 @@ def check_column_spoilers(cand_stars, ok, stars, n_sigma):
     :returns: bool mask on cand_stars marking column spoiled stars
     """
     column_spoiled = np.zeros_like(ok)
+    rej = []
     for idx, cand in enumerate(cand_stars):
         if not ok[idx]:
             continue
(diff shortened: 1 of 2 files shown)


proseco/guide.py
```diff
@@ -453,13 +558,21 @@ def get_ax_range(r, extent):
...
         pix = np.array(dark.aca[rminus:rplus, cminus:cplus])
-        pixmax = max(np.max(bin2x2(pix)),
-                     np.max(bin2x2(pix[1:-1])),
-                     np.max(bin2x2(pix[:, 1:-1])),
-                     np.max(bin2x2(pix[1:-1, 1:-1])))
+        pixmax = 0
+        max_r = None
+        max_c = None
+        for i in range(0, pix.shape[0] - 1):
+            for j in range(0, pix.shape[1] - 1):
```
jeanconn: This double-loop over the pixels is going to be slower than bin2x2; I just didn't have time to figure out the 4 different indexes into the 2x2 bins sums if I wanted the pixel location of the warmest pixel 2x2.

proseco/guide.py
```diff
@@ -84,6 +84,14 @@ class GuideTable(ACACatalogTable):
 
     dark = MetaAttribute()
     cand_guides = MetaAttribute(is_kwarg=False)
+    reject_info = MetaAttribute(default=[], is_kwarg=False)
+
+    def reject(self, reject):
+        """
+        Add a reject dict to self.reject_info
+        """
+        reject_info = self.reject_info
```
taldcroft: This one-off is not helping anything.

proseco/guide.py
taldcroft: Can you do timing for a few obsids with and without this PR?  Just make sure it isn't totally blowing up.

proseco/tests/test_guide.py
```diff
@@ -72,7 +72,8 @@ def test_box_mag_spoiler():
     assert 688523960 in selected1['id']
 
     # Set the spoiler to be 10th mag and closer to the second star
-    stars.get_id(688522000)['mag'] = 10.0
+    stars.get_id(688522000)['mag'] = 9
+    stars.get_id(688522000)['row'] = 50
```
taldcroft: Why these changes?  I thought all updates were strictly about adding debug info.

jeanconn: As far as I could tell, this test was no longer actually testing what it said it was.  Didn't notice until more debug info.  The original comment mentioned moving the star closer which seems to have gotten cut at some point, and the mag difference wasn't actually enough to hit the box test either.

proseco/guide.py
```diff
@@ -337,8 +412,15 @@ def check_spoil_contrib(cand_stars, ok, stars, regfrac, bgthresh):
             val = on_region[pixlabel == chandra_aca.aca_image.EIGHT_LABELS][0]
             if val > bgthresh:
                 bg_spoiled[cand_stars['id'] == cand['id']] = True
+                rej.append({'id': cand['id'],
+                            'type': 'region background spoiled',
+                            'bg_thresh': bgthresh,
+                            'bg_pix_val': val,
+                            'pix_id': pixlabel,
+                            'text': f'Cand {cand["id"]} has bg pix spoiled by spoilers'})
+                continue
```
jeanconn: This "continue" is another functional change.  If one background pixel is spoiled there's no need to check the rest of them.

proseco/guide.py
jeanconn: It looks like the densest field takes 20% longer with this PR (but I didn't isolate this code).  Other fields look 5 to 10% longer.  Is that "totally blowing up"?

Should I just add the more verbose logging of what happened to all the candidates as an option?  It isn't clear to me how much we need this output all the time and how much we could just generate it later from the archived inputs.

jeanconn: I think I've fixed the binned max code so there's no performance impact.  Will look at the pickle sizes next.

jeanconn:
With regard to pickle size, it looks to me like this increases the guide portion of the pickle by ~5x (on a small sampling), which increases, on average, the overall pickle size by 10%.  I'll look at the dense fields.

I was also looking at not stopping the stages when enough stars were found (which will increase time and size).

taldcroft:
About the not-stopping option, maybe that is best left for a debug mode.  From what you've said this PR (in present state) seems fine to me.

PR #138, opened 2018-09-29
Update merged catalog order to match matlab tools

### LINKED ISSUE #137: idx slot mismatches vs matlab tools
It looks like the matlab tools right now are putting the selected stars in order by AGASC ID within a type (instead of mag or supplied order or ...)

It is probably easiest to just do this also on the proseco side for the top level catalog.

jeanconn:
Update merged catalog order to match matlab tools

Closes #137


proseco/catalog.py
```diff
@@ -316,9 +316,9 @@ def merge_cats(fids=None, guides=None, acqs=None):
     def filt(tbl, id_set):
         return [row['id'] for row in tbl if row['id'] in id_set]
 
-    bot_ids = filt(guides, set(guides['id']) & set(acqs['id']))
-    gui_ids = filt(guides, set(guides['id']) - set(acqs['id']))
-    acq_ids = filt(acqs, set(acqs['id']) - set(guides['id']))
+    bot_ids = sorted(filt(guides, set(guides['id']) & set(acqs['id'])))
```
taldcroft: Put the `sorted` into the `filt` function (with additional comment about sorting by ID within each group to match what Matlab does).

taldcroft:
Obsid 20848 has GUI stars that are not sorted by mag or ID.  So not clear what's happening or where to go with this.
```
---------------------------------------------------------------------------------------------
 IDX SLOT        ID  TYPE   SZ   P_ACQ    MAG   MAXMAG   YANG   ZANG DIM RES HALFW PASS NOTES
---------------------------------------------------------------------------------------------
[ 1]  0           2   FID  8x8     ---   7.000   8.000   -773  -1741   1   1   25          
[ 2]  1           4   FID  8x8     ---   7.000   8.000   2140    166   1   1   25          
[ 3]  2           5   FID  8x8     ---   7.000   8.000  -1826    160   1   1   25          
[ 4]  3   414324232   BOT  6x6   0.985   8.242   9.750  -2260    915  12   1   80          
[ 5]  4   413805792   GUI  6x6     ---   7.286   8.781   2356     -7   1   1   25          
[ 6]  5   413803744   GUI  6x6     ---   8.868  10.375   2193   2256   1   1   25          
[ 7]  6   414322472   GUI  6x6     ---   8.707  10.203  -2143  -1790   1   1   25         C
[ 8]  7   414326136   GUI  6x6     ---   7.331   8.828   -151    535   1   1   25          
[ 9]  4   414330208   ACQ  6x6   0.985   6.518   8.016  -2057   1512  28   1  160    a2    
[10]  5   413799128   ACQ  6x6   0.985   8.697  10.203   1349    150  28   1  160    a2    
[11]  6   414329112   ACQ  6x6   0.970   9.195  10.703    795   -687  28   1  160    a2    
[12]  7   414329664   ACQ  6x6   0.963   9.274  10.781    738    -67  28   1  160    a2    
[13]  0   414324712   ACQ  6x6   0.951   9.399  10.906  -1073   -664  28   1  160    a2    
[14]  1   414327264   ACQ  6x6   0.879   9.405  10.906   -570    924  28   1  160    a2    
[15]  2   413798376   ACQ  6x6   0.868   9.440  10.938   1540    824  28   1  160    a2   C
```

jeanconn:
Yeah, this is still somewhat WIP, but I think that the way the our tools our integrated it will end up in AGASC ID order.  I think if our merged cat matches that's fine.

My plan was to check via integration testing and evaluation of the pickle.

Regarding the integration, this line is basically chucking our order and putting the acq and guide catalogs in agasc id order (this was what I thought I might fix)

https://github.com/sot/fot_matlab_tools/blob/bac55a7f9695bd36599e4ddca0cb04ced435c4ae/MissionScheduling/stars/StarSelector/%40StarSelector/private/assignProsecoCatalog.m#L68

I think that this line when matlab builds the table ends up sorting the BOTS.  It may not sort the acqs or guides but if those are already sorted by agasc id from the line above I think it will persist.  That's what I need to check.

https://github.com/sot/fot_matlab_tools/blob/review_4.0/MissionScheduling/stars/%40ACACatalog/private/convertStarSelectorClass.m#L35

jeanconn:
After puzzling more in the matlab code, I think we'll end up with a different compromise on the sorting anyway (sorting the BOTs by ID, trying to keep the proseco-supplied order for plain GUI and ACQ) so the regression test update would need redo anywa.

jeanconn:
I think this is good to go but won't be functionally verified until the change is merged on the matlab side and end-to-end tested.

proseco/catalog.py
jeanconn: Since now it is just sorting the bots, I don't think it makes sense to push into `filt`.

PR #139, opened 2018-09-29
Script to run starcheck on proseco-generated catalogs

taldcroft:
This incorporates the script I used to generate products for the catalogs shown in StarWorkingGroupMeeting2018x07x18.  

To do:

- [x] The merge catalogs function is redundant with what is in `catalog.py` now, so probably take it out.
- [x] Deal with guide and fid which now come from proseco also.
- [x] @jeanconn apparently wrote something similar, so reconcile the two.

The new `validate` is a fine place to dump notebooks in progress and put on a branch for visibility on github.  Most likely will not merge though.

### DIFF
--- a/validate/starcheck_proseco.py
+++ b/validate/starcheck_proseco.py
@@ -1,3 +1,42 @@
+"""
+Do three steps to run starcheck on proseco-generated catalogs for one or more flight weekly loads:
+
+- Copy the flight load products into a local working dir
+- Generate proseco catalogs and update products in place to reflect the new cats
+- Run starcheck
+
+It is assumed the flight load products are in a structure like::
+
+  <load_root>/2018/JUL0218/oflsa  <==  For <load_name> = JUL0218A
+
+The outputs go into a directory like::
+
+  <out_root>/JUL0218A/
+
+Typically this is going to be run from the proseco git repo, using::
+
+  $ ska3
+  $ cd ~/git/proseco
+  $ git checkout <branch/tag>  # If doing run-for-record, be sure to use the release tag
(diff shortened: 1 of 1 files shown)


jeanconn:
For merging, I've got some code that subs in guide stars, but at this point, do we just want a new code that subs in the merged catalog from get_aca_catalog anyway?  Or independent substitutions for different kinds of tests?

taldcroft:
I'm planning to work on this at a high priority so we can have proseco load products for your starcheck work.

taldcroft:
Running this now on HEAD for the canonical 10 weekly loads.  My goodness it is slow on kadi.

taldcroft:
I've added links to all the starcheck results on the Proseco TWiki special projects page.  Getting clean output from starcheck will either be a lot of work or require more trust in proseco.  Hmm.

jeanconn:
And we foresaw the issue that we're using different definitions of bad pixels at this point. I think for starcheck it makes sense to reimplement the "single bad pixel in region" check with a relative mag threshold equivalent to something like 10.3 mags : 2000 e-/sec .  Or reimplement it with the 2x2 bins (I could go either way on what we consider more conservative).  Could also chuck the bad pixel detection altogether I think if we run annie-lite on all observations in review.

taldcroft:
Oops, forgot to do the final rsync after it all finished.  Should be good now.

taldcroft:
I'm not planning to stub in FIDSEL right now.  Could be a cleanup later.  I think that Matlab integration testing in the super-schedule etc will test fids sufficiently.

PR #141, opened 2018-09-29
Avoid bad pixels and bad stars in acq and fid selection

### LINKED ISSUE #34: Avoid bona fide bad pixels and bad stars
There are some actual bad (not just hot) pixels / bad pixel regions that should be avoided.

This applies to fid, guide and acq selection.

taldcroft:
Closes #34


taldcroft:
Yes. I don't know where it is though.

jeanconn:
I'm not sure how much padding we want on it.

https://github.com/sot/fot_matlab_tools/blob/review_4.0/Characteristics/characteristics_Stars.m#L192
http://cxc.cfa.harvard.edu/mta/ASPECT/aca_weird_pixels/

taldcroft:
The related issue of avoiding configurations that spoil a fid due to the fid trap at (-374,347) is now tracked in #142.

PR #146, opened 2018-09-30
Exclude stars that trigger bad fid pixel

### LINKED ISSUE #142: Avoid guide stars that trigger the bad pixel for ACIS-I-6
Based on review of http://cxc.cfa.harvard.edu/mta/ASPECT/aca_weird_pixels/.  Here is a proseco catalog plot to illustrate, where fid idx=3 has the problem and acq idx=11 was forced in and represents the star that causes the readout timing issue.

image

Since fid selection happens before guide, for 4.0 we just take the fids as immutable for guide selection and reject guide candidates which have (for any fid):
```
bad_row = -374
bad_col = 347
fid_margin = 8  # 20" fid positional uncertainty + 4 pixel readout window
abs(fid['col'] - bad_col) < fid_margin  # AND
abs(fid['row'] - abs(guide['row']) - bad_row) < fid_margin + dither / 5
```

The ACIS-I-6 fid has nominal position row, col = (-71, 346).  So this will ding guide stars that have `abs(row)` close to 303.  In obsid 1576, guide star 367657896 has row=302.

This is all probably a repeat of existing code in Matlab and/or starcheck, but I wanted to think about it ab-initio.

jeanconn:
Exclude stars that trigger bad fix pixel

See http://cxc.cfa.harvard.edu/mta/ASPECT/aca_weird_pixels/

Closes #142


proseco/guide.py
```diff
@@ -64,17 +65,18 @@ class GuideTable(ACACatalogTable):
...
+    yaml_exclude = ('stars', 'cand_guides', 'dark', 'known_fids')
+    pickle_exclude = ('stars', 'dark', 'known_fids')
 
     # Name of table.  Use to define default file names where applicable.
     # (e.g. `obs19387/guide.yaml`).
     name = 'guide'
 
     # Required attributes
-    required_attrs = ('att', 't_ccd_guide', 'date', 'dither_guide')
+    required_attrs = ('att', 't_ccd_guide', 'date', 'dither_guide', 'n_guide')
```
jeanconn: Adding n_guide to required_attrs snuck in here, but I think it can just stay here and go with this PR unless there are objections.

proseco/guide.py
```diff
@@ -30,6 +30,7 @@ def get_guide_catalog(obsid=0, **kwargs):
     :param date: date of acquisition (any DateTime-compatible format)
     :param dither: dither size 2-element tuple: (dither_y, dither_z) (float, arcsec)
     :param n_guide: number of guide stars to attempt to get
+    :param known_fids: selected fids (used for guide star exclusion)
```
taldcroft: You don't need to invent something new here.  The `fids` property will already handle it, so the only change to have the `GuideTable` know about fids is to call `get_guide_catalog(stars=stars, fids=aca.fids, **kwargs)`.  Look at `acq.py` for a model of how that works.

There should not even be an issue in pickling / garbage collection because there is no circular reference (guide => fid => guide).  Also, my understanding is that pickling will only store an object once and otherwise just store refs to that object.

proseco/guide.py
```diff
@@ -340,6 +349,45 @@ def get_initial_guide_candidates(self):
...
+    bad_row = GUIDE_CHAR.fid_trap['row']
+    bad_col= GUIDE_CHAR.fid_trap['col']
+    fid_margin = GUIDE_CHAR.fid_trap['margin']
+
+    # Check to see if the fid is in the zone that's a problem for the trap and if there's
+    # a star that can cause the effect in the readout regiser
+    for fid in fids:
+        incol = abs(fid['col'] - bad_col) < fid_margin
+        inrow = (fid['row'] < 0) & (fid['row'] > bad_row)
+        if (incol & inrow):
```
taldcroft: Should be `if incol and inrow:`  (not a bitwise `&` but rather a logical `and`).  And no extraneous parens for simple expressions like this.

proseco/guide.py
jeanconn: I saw all kinds of stuff related to fids in `acq.py` but still haven't figured out what to do to just get it to be an allowed keyword arg for GuideTable.

proseco/guide.py
```diff
@@ -340,6 +349,45 @@ def get_initial_guide_candidates(self):
...
+            fid2trap = fid['row'] - bad_row
+            star2reg = 512 - abs(cand_stars['row'])
```
taldcroft: Either add code comments or give these descriptive names like `fid_dist_to_trap` and `star_dist_to_register`.

proseco/guide.py
```diff
@@ -340,6 +349,45 @@ def get_initial_guide_candidates(self):
         return cand_guides
 
 
+
+def check_fid_trap(cand_stars, fids, dither):
+    """
+    Search for guide stars that would cause the fid trap issue and mark as spoilers.
```
taldcroft: Add :params: and :returns: to docstring.

proseco/guide.py
```diff
@@ -340,6 +349,45 @@ def get_initial_guide_candidates(self):
...
+    """
+
+    spoilers = np.zeros(len(cand_stars)).astype(bool)
+    rej = []
+
+    if len(fids) == 0:
+        return spoilers, []
+
+    if not isinstance(dither, ACABox):
```
taldcroft: At this point in the code `dither` must be an `ACABox`, right?  This function is not really called as a standalone thing and you can specify in the docstring that is has to an `ACABox`.

proseco/guide.py
```diff
@@ -340,6 +349,45 @@ def get_initial_guide_candidates(self):
         return cand_guides
 
 
+
+def check_fid_trap(cand_stars, fids, dither):
```
taldcroft: It looks like there is precedent, but I don't see why you change from `cand_guides` to `cand_stars` for the arg name.  `cand_guides` is the name of the thing you use when calling, so when reading this code I got confused, wondering if there is a different concept of candidate *stars*.  It helps code clarity (for your readers) to be consistent with naming.

taldcroft: And at some level these should all just become methods of `GuideTable` and then you need no args at all.  Not a requirement for 4.0.

proseco/guide.py
jeanconn: If I try to just supply it and I get

```
-> 101     aca.guides = get_guide_catalog(stars=stars, fids=aca.fids, **kwargs)
    102 
    103     # Make a merged starcheck-like catalog.  Catch any errors at this point to avoid

/proj/sot/ska/jeanproj/git/proseco/proseco/guide.py in get_guide_catalog(obsid, **kwargs)
     39 
     40     guides = GuideTable()
---> 41     guides.set_attrs_from_kwargs(obsid=obsid, **kwargs)
     42     guides.set_stars()
     43 

/proj/sot/ska/jeanproj/git/proseco/proseco/core.py in set_attrs_from_kwargs(self, **kwargs)
    222                 setattr(self, name, val)
    223             else:
--> 224                 raise ValueError(f'unexpected keyword argument "{name}"')
    225 
    226         # If an explicit obsid is not provided to all getting parameters via mica

ValueError: unexpected keyword argument "fids"

```

It thought it would be set as an allowed keyword arg on 'guide' by MetaAttribute.__set_name__ but that didn't seem to get called and I couldn't figure it out.

jeanconn:
Taking a break. Will do cleanup on these items later today.

proseco/guide.py
taldcroft: You need this, which you can put into this PR:
```
diff --git a/proseco/guide.py b/proseco/guide.py
index f1e027f..198d614 100644
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -71,6 +71,8 @@ class GuideTable(ACACatalogTable):
     # (e.g. `obs19387/guide.yaml`).
     name = 'guide'
 
+    allowed_kwargs = ACACatalogTable.allowed_kwargs | set(['fids'])
+
     # Required attributes
     required_attrs = ('att', 't_ccd_guide', 'date', 'dither_guide')
```

proseco/guide.py
taldcroft: I started there too, but eventually in many cases the benefit of using class methods became obvious.  For instance you could call `self.reject` (or whatever that is called) or `self.log` and thus clean up the API alot by not having those `rej` return args.

FYI the log machinery was designed to allow capture of anything in the `data` kwarg (e.g. those reject dicts), not just text messages.  Then you automatically get the function name and time where a rejection happened (FWIW).  You can toss in extra kwargs like `reject=True`.  This is how warnings are handled.

In practice you aren't doing independent unit tests now (for this particular function), and it is easy enough to do so by just making an empty `GuideTable()` and setting attributes.  See e.g.
https://github.com/sot/proseco/blob/8db4fa243b68d5744f3347b8ce08891e9c127372/proseco/tests/test_acq.py#L31

proseco/characteristics_guide.py
```diff
@@ -1,5 +1,10 @@
 import numpy as np
 
+# Fid trap effect
```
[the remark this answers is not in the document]
jeanconn: This is done.

proseco/guide.py
jeanconn: This is not pretty but done

proseco/guide.py
jeanconn: OK.  I'll plan to clean up `cand_stars` at some point  (I think they were all `cand_stars` when I started but you liked `cand_guides` better and then I just didn't implement it universally) and move the reject crap into logging, but I think this is functional at this point.

PR #147, opened 2018-09-30
Excise YAML references and support code

taldcroft:
2-minute PR to clean out all the code related to YAML serialization.  There were no remaining tests of this and it is just totally broken at this point.  If we ever need it again this is here in the history to revert.

Will wait until after #146 since this would probably create a merge conflict in that.


jeanconn:
I've just compressed my commits in #146 and removed the change that fiddled with yaml_exclude, so I think this can go whenever.

taldcroft:
Excellent.

PR #165, opened 2018-10-05
Add calc_spoiler_impact function and use instead of box filter for acq

taldcroft:
This computes the centroid offset and fractional change in summed counts caused by spoilers.
It runs in about 1 msec for a single spoiler. 

To do:
- [x] Unit tests with synthetic values to demonstrate expected behavior
- [x] Add 21068 functional test
- [x] Fix current breaking test (include_exclude).
- [x] Code comments

### DIFF
--- a/proseco/acq.py
+++ b/proseco/acq.py
@@ -18,7 +18,7 @@ from chandra_aca.transform import (pixels_to_yagzag, mag_to_count_rate,
 from . import characteristics as CHAR
 from .core import (get_mag_std, ACACatalogTable, bin2x2,
                    get_image_props, pea_reject_image, ACABox,
-                   MetaAttribute)
+                   MetaAttribute, calc_spoiler_impact)
 
 
 def get_acq_catalog(obsid=0, **kwargs):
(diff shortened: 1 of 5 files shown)


taldcroft:
The basic idea is the same as for guide, details slightly different.

jeanconn:
Interesting.  I had played with using the fm centroid, but didn't have this idea of using a mean high background estimate. 

There are the couple of edge cases that I think of wrt to this approach... that multiple spoilers can "balance out" (should be rare and we don't care for acquisition) or that the spoiling star right on the edge of the 6x6 ends up coming more onto that pixel than modeled here.  For acquisition I didn't know if that second case might be an issue, because I thought the placement of the 8x8 is hits-based not coordinate based.

But probably all in the noise.

jeanconn:
> On the issue of critical behavior near a pixel boundary, the plots in the notebook show that there does not appear to be very strong pixel-related changes in the output values.

Gotcha.  There's not much text there to hold my hand through what you've done to validate that, so it is going to take me longer to understand.

taldcroft:
If you look at the images showing centroid offsets, they are relatively smoothly varying and you don't see strong square-like structures corresponding to pixel boundaries.  That's all I meant.

taldcroft:
To be clear you can see evidence of the pixels, but not like a wall somewhere.

taldcroft:
For use with acq selection, I think this is good and I'm merging.

PR #167, opened 2018-10-08
WIP: annie-lite guide simulator

taldcroft:
This notebook shows a proof of concept for an annie-lite guide simulator for evaluating a guide catalog.  This strips out much of the "high-fidelity" part of annie and instead just tracks stars without the real PEA and OBC behavior that slows down annie.  Note that it does not do the closed-loop OBC PCAD control, but just does a perfect dither pattern.

First look with obsid 21068 seems to give qualitatively OK results.

Speed is encouraging, about 1.6 secs for 2 ksec at 2.05 s/readout.  This is with absolutely no attempt to optimize, and I am hopeful there is low-hanging fruit in ACAImage.


taldcroft:
This can now do 1000 readout frames in about 0.7 sec.

taldcroft:
@malgosias - have you taken the code here and incorporated it into some other home that can be the new baseline for annie-lite?  If so, can you give the reference here.  If not, can you make it so?

I want to close out this PR which is just noise here in proseco at this point.

PR #168, opened 2018-10-11
Add WIP notebook evaluating proseco guide cat performance

jeanconn:


jeanconn:
This should probably go in the validate directory.  And I should probably replace annie and my hack to just halve the dither period with with annie-lite and see how it goes.

taldcroft:
Bringing comments on the validation notebook here.

taldcroft:
Questions to answer:
- How many catalogs are in the sample?
- Related to cell 12, what fraction of flight catalogs have max_extent < 2500?
- Similarly, Eric seems most concerned by all stars on one quadrant.  How do those stats compare?

taldcroft:
Can you do comparisons of `n_stars` using the new piecewise linear metric?  Hopefully that is solidly in favor of proseco.

taldcroft:
In cell 17 (comparing dr95), it would be useful to see a scatter plot of the two values along side the histogram of diffs.

The positive distribution up to 50 (and mostly just up to about 30) is OK.

The two positive outliers (especially proseco - as-run = 100 arcsec) need comment.  The question is why that can happen?  This analysis is (correctly) doing apples-to-apples (both annie sims) and proseco guide selection is explicitly aware of warm/hot pixels while SAUSAGE did a blanket avoidance of pixels in a list that was not usually up to date.

jeanconn:
Regarding "The two positive outliers (especially proseco - as-run = 100 arcsec) need comment. The question is why that can happen? This analysis is (correctly) doing apples-to-apples (both annie sims) and proseco guide selection is explicitly aware of warm/hot pixels while SAUSAGE did a blanket avoidance of pixels in a list that was not usually up to date." I have had this question but I don't know how you'd like to answer it.   As far as I can tell, this just speaks to the fact that the non-linearities of the scenario of warm and hot pixels around a constellation of tracked guide stars can end up with worse tracking performance in the simulation even if the individual stars are probably better.  But annie doesn't really have a way to tell me how and why.   Or proseco guide star selection has a bug.  Or annie is just wrong.

jeanconn:
Regarding "How many catalogs are in the sample?" That's mostly in the last cell but I can itemize with the count of catalogs excluded because they had manual edits etc.

taldcroft:
With annie you can get the observed minus true star centroids for each slot.  To get an extreme outlier in dr95 you must have at least one slot that is quite bad for some reason.  Chasing this down will hopefully teach us *something*.  What that something is I cannot predict.

jeanconn:
"To get an extreme outlier in dr95 you must have at least one slot that is quite bad for some reason."

Well, and I referenced the min and maximum individual star residuals in this notebook.  I'm just not as confident that the trends work out the way you hope (I think you can see high dr95 without this "red flag" bad star), but I'll look at these two outliers (probably again).

jeanconn:
Also, not sure if I should copy these comments into #223 as the notebook in this PR is largely unrelated to the comments.

jeanconn:
Here is a look at one of the outliers, obsid 19329.  

https://nbviewer.jupyter.org/url/cxc.harvard.edu/mta/ASPECT/proseco/proseco_guide_outlier_19329.ipynb

It looks to me like all the star residuals are comparable between the proseco and SAUSAGE star selections, but from a look at the catalogs the proseco guide selected stars are going to end up in a smaller box.  So I think this is a case where the figure of merit optimization in SAUSAGE had a side effect of setting things up to have less roll error.  But annie doesn't really help me to get to that conclusion.  I think we'd have to re-implement the figure of merit calculation to even run the numbers.

jeanconn:
Somewhat equivalently, I expect that the proseco guide star selection would select the same stars as SAUSAGE with somewhat relaxed thresholds on the local dark map; this falls more under the category of parameter optimization of the existing states.

taldcroft:
> did you want old dr95th vs new dr95th or some such?)

Bingo.

jeanconn:
OK.  I've just avoided it because it isn't particularly illuminating.

taldcroft:
I assume this can be closed now?

PR #170, opened 2018-10-11
Add option to update_products to use proseco pkl file for cats

jeanconn:
Add option to update_products to use proseco pkl file for cats

The idea is that the pkl file from ORViewer could be used as the source for the stubbing in of catalogs.  From ORViewer I also already have the updated DOT and guide summary, but am ignoring them for now.

This doesn't immediately work with the "run_all" idea, and is designed for manual calls to update_products.

### DIFF
--- a/validate/starcheck_proseco.py
+++ b/validate/starcheck_proseco.py
@@ -40,6 +40,7 @@ or remove:
 from pathlib import Path
 import subprocess
 import shutil
+import pickle
 
 import parse_cm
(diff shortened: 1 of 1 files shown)


jeanconn:
I should set this up so the "run_all" behavior goes ahead and makes a reasonable pkl of the proseco catalogs it just created and inserted.

taldcroft:
This generally looks OK.  You should add some words in the module doc string about this usage.

PR #176, opened 2018-12-01
Implement include/exclude_ids in guide star selection

jeanconn:
Implement include/exclude_ids in guide star selection

### DIFF
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -117,6 +117,10 @@ class GuideTable(ACACatalogTable):
             # cand_guides as the 'selected' stars.
             return cand_guides
         cand_guides['stage'] = -1
+        # Force stars in include_ids to be selected at stage 0
+        for star_id in self.include_ids:
+            cand_guides['stage'][cand_guides['id'] == star_id] = 0
+            self.log(f'{star_id} selected in stage 0 via include_ids', level=1)
         n_guide = self.n_guide
         for idx, stage in enumerate(GUIDE_CHAR.stages, 1):
             already_selected = np.count_nonzero(cand_guides['stage'] != -1)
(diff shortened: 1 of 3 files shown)


jeanconn:
This was the most direct hack to accomplish the behavior.  I didn't have in mind this use case when doing the original implementation, so it could really use a rewrite to have functions to manipulate the list of candidate stars.

proseco/guide.py
```diff
@@ -261,53 +266,84 @@ def get_initial_guide_candidates(self):
         stars = self.stars
         dark = self.dark
 
-        # Use the primary selection filter from acq, but allow bad color
-        # and limit to brighter stars
-        ok = ((stars['CLASS'] == 0) &
```
taldcroft: A corollary to the recommendation of using `ok` for boolean masks is that this should only be done for code that is immediately adjacent.  So move this statement down to [where it is used](https://github.com/sot/proseco/blob/1c14d0e1926aa105aa243b72ec3606d9bb7794ea/proseco/guide.py#L289).

jeanconn:
Thanks @taldcroft .  
-  Forgot to mark as WIP
-  I think you are basically right about implementation
-  My pre-this-hack version was a bit more like what you are describing. But since I annotate the larger set of candidates with information about how they behaved in the checks somewhat as a side effect of running those checks,  my brain was not happy with not having values for those checks for a forced candidate.  Thus, this first full implementation carried along the forced candidates and put them through the checks (to have more complete information for the report).  However, for the special case of forced stars it probably makes more sense to just handle debug and warnings from the starcheck side.

jeanconn:
I've just updated this to be more consistent with @taldcroft 's vision.  There are still compromises in some code that should be addressed in guide star selection, but this is at least a more lean change.

jeanconn:
I ended up not adding back in code to check that include_ids aren't in exclude_ids.

proseco/tests/test_catalog.py
```diff
@@ -36,7 +36,7 @@ def test_get_aca_catalog_20603():
     """Put it all together.  Regression test for selected stars.
     """
     # Force not using a bright star so there is a GUI-only (not BOT) star
-    aca = get_aca_catalog(20603, exclude_ids=[40113544], n_fid=2, n_guide=6, n_acq=7,
+    aca = get_aca_catalog(20603, exclude_ids_acq=[40113544], n_fid=2, n_guide=6, n_acq=7,
```
taldcroft: Related to the discussion in #178, maybe revert this test and then fix the expected catalog to show both guide and acq exclusion of ..3544?

taldcroft: OK fair enough.

PR #179, opened 2018-12-05
Lower limit of mag_err=0.3 for color=1.5 stars

taldcroft:
For color=1.5 stars set a lower limit of 0.3 on the catalog mag error. This is based on analysis in ipynb/star_selection/star-mag-std-dev.ipynb which shows that for color=1.5 stars the observed mag error is does not correlate well with mag_err below around 0.3 (on the negative side) and so capping at 0.3 is a reasonable way to capture the uncertainty.

image

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -897,8 +897,8 @@ class StarsTable(ACACatalogTable):
         # For color=1.5 stars set a lower limit of 0.3 on the catalog mag error.
         # This is based on analysis in ipynb/star_selection/star-mag-std-dev.ipynb
         # which shows that for color=1.5 stars the observed mag error is
-        # does not correlate well with mag_err below around 0.3 and so capping
-        # at 0.3 is a reasonable way to capture the uncertainty.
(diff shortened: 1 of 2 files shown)


proseco/core.py
```diff
@@ -893,6 +893,15 @@ def from_stars(cls, att, stars, logger=None, copy=True):
         # the true star mag) and the sample uncertainty in the ACA readout mag
         # for a star with mag=MAG_ACA.  The latter typically dominates above 9th mag.
         mag_aca_err = stars['MAG_ACA_ERR'] * 0.01
+
+        # For color=1.5 stars set a lower limit of 0.3 on the catalog mag error.
+        # This is based on analysis in ipynb/star_selection/star-mag-std-dev.ipynb
+        # which shows that for color=1.5 stars the observed mag error is
+        # does not correlate well with mag_err below around 0.3 and so capping
```
jeanconn: I'll just nick pick in the comment and say that I've only heard "capping" used for setting upper bounds.  I don't think there is a equivalent for setting a lower limit (floor-ing?).

PR #184, opened 2018-12-06
Tweak include/exclude args to be lists of ints as needed

jeanconn:
Tweak include/exclude args to be lists of ints as needed

The include_*, exclude_* kwargs seems to end up as either empty numpy arrays, numpy arrays of floats, or single numpy ints when passed to Python from Matlab with pyexec.  I think none of those work in get_aca_catalog.  This PR is intended to cast all of them to lists of ints.

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -2,6 +2,7 @@ import functools
 import pickle
 import inspect
 import time
+import collections
 from copy import copy
 from pathlib import Path
 
@@ -371,11 +372,23 @@ class ACACatalogTable(Table):
             self._default_formats[name] = '.6f'
 
     def set_attrs_from_kwargs(self, **kwargs):
+
+        # The include/exclude kwargs appear to need to be undefined or lists of ints
+        for name in ['include_ids_acq', 'include_halfws_acq', 'include_ids_guide',
+                    'exclude_ids_acq', 'exclude_ids_guide']:
+            if name in kwargs:
+                if not isinstance(kwargs[name], collections.Iterable):
+                    kwargs[name] = [int(kwargs[name])]
+                else:
+                    kwargs[name] = [int(i) for i in kwargs[name]]
+
+
         for name, val in kwargs.items():
             if name in self.allowed_kwargs:
                 setattr(self, name, val)
             else:
                 raise ValueError(f'unexpected keyword argument "{name}"')
+
 
         # If an explicit obsid is not provided to all getting parameters via mica
         # then all other params must be supplied.


jeanconn:
@taldcroft Should be able to come up with something nicer that works with the attribute syntax.  I just didn't see a good way to deal with this on the Matlab side, and also did not want to write Python subroutines to munge the kwargs as lambda functions in the Matlab.

proseco/core.py
```diff
@@ -371,12 +372,24 @@ def __init__(self, data=None, **kwargs):
...
+                    kwargs[name] = [int(i) for i in kwargs[name]]
+
+
         for name, val in kwargs.items():
             if name in self.allowed_kwargs:
                 setattr(self, name, val)
             else:
                 raise ValueError(f'unexpected keyword argument "{name}"')
 
+
```
jeanconn: Oops.

taldcroft:
And then make a subclass for those attributes:
```
class Int1dMetaAttribute(MetaAttribute):
    def __set__(self, instance, value):
        instance.meta[self.name] = np.atleast_1d(value).astype(int)
```

jeanconn:
Yes, np.atleast_1d seems to help just fine.  Per my comment in the email thread, I think this can just go in the updated Matlab file and doesn't need a proseco edit.

taldcroft:
Er. I'd prefer to make the Python accept different args in a generalized way rather than hacking the Matlab Python calls.  I'm guessing you are going to end up with a lot of repetition in the Matlab/Python calls?

jeanconn:
OK.  I didn't see much value added on the Python side and was just trying to avoid yet another release.  On the matlab side it was just looking like:
```
get_aca_catalog(...
include_ids_acq = np.atleast_1d($include_ids_acq), include_ids_guide=np.atleast_1d($include_ids_guide), ..
```
So yeah, you repeat np.atleast_1d 5 times.

taldcroft:
That's not fixing the type to int.  I suspect that getting floats will break things, though maybe you've already tested?

jeanconn:
Yes, I was live testing the np.atleast_1d idea.  I think the "or floats" was a mistake on my part in reading a stack trace and they were still coming through as numpy ints.

jeanconn:
Though now failing in a different area.

```
In [16]: %debug
> /proj/sot/ska/jeanproj/git/proseco/proseco/acq.py(498)select_best_p_acqs()
    496                 if accepted:
    497                     acq_indices.append(acq_idx)
--> 498                     box_sizes.append(box_size)
    499 
    500                 if len(acq_indices) == self.n_acq:

ipdb> box_size
160
ipdb> box_sizes
array([], dtype=float64)
```

jeanconn:
Right, so do we care about this for 4.3 or is the matlab sufficient
```
            'include_ids_acq=list(np.atleast_1d($include_ids_acq)),'...
            'exclude_ids_acq=list(np.atleast_1d($exclude_ids_acq)),'...
            'include_halfws_acq=list(np.atleast_1d($include_halfws_acq)),'...
            'include_ids_guide=list(np.atleast_1d($include_ids_guide)), '...
            'exclude_ids_guide=list(np.atleast_1d($exclude_ids_guide)))'];
```

taldcroft:
See #187.

taldcroft:
If we get in a time crunch then the above makes me unhappy but works.

jeanconn:
Define "time crunch".

PR #187, opened 2018-12-07
Add IntListMetaAttribute for include/exclude attrs

taldcroft:
If possible let's go with this for 4.3.  Just need to add a test from get_aca_catalog but I expect it to work based on new unit tests.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -51,16 +51,16 @@ focus_offset  SIM focus offset [steps] (default=0)
 
 **Optional**
 
-=================== =========================================================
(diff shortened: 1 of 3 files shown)


jeanconn:
I was planning to try this by making tests first, it looks like with the plain pyexec pass through you get an int for a single value, but a numpy array for an empty list (empty ones apparently of type float64) or multiple values.  I think that will work with your code but isn't consistent with the doc.

jeanconn:
Looks OK against NOV2618 with ORViewer.  Merging.

Issue #125, opened 2018-09-26
Add more specific tests of guide spoilers


taldcroft:
The #105 merge showed a gap in testing for guide spoilers.  Add tests with a synthetic catalog(s) where the result can be analytically known in advance.

taldcroft:
Since we believe the guide spoiler handling is correct, I've pushed this off to 4.1.

taldcroft:
Done with #218.

PR #195, opened 2018-12-14
Fix bug that was only using row dither for hot pixel region check

### LINKED ISSUE #194: Guide star selection: pixel region not calculated correctly for asymmetric dither
These bits in this subroutine should reference "extent" not "row_extent"

https://github.com/sot/proseco/blob/52145e0993239c55dd40d75046d9f6c4ce9a8b57/proseco/guide.py#L622

jeanconn:
Fix bug that was only using row dither for hot pixel region check

`get_ax_range` was ignoring the supplied value of "extent" and using row_extent always.  This is an issue for observations with asymmetric dither.

`get_ax_range` is still ugly, but I've moved it to top level so I can use it equivalently from the guide reporting.

Closes #194

### DIFF
--- a/proseco/characteristics_guide.py
+++ b/proseco/characteristics_guide.py
@@ -9,6 +9,9 @@ fid_trap = {'row': -374,
 # Minimum scaled guide count for thumbs_up
 min_guide_count = 4.0
 
+# Add this padding to region checked for bad pixels (in addition to dither)
+dither_pix_pad = 0.4
+
 CCD = {'row_min': -512.0,
        'row_max': 512.0,
        'col_min': -512.0,
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -665,7 +665,9 @@ def get_imposter_mags(cand_stars, dark, dither):
                 idx = np.unravel_index(np.argmax(bin_image), bin_image.shape)
                 max_r = rminus + row_off + idx[0] * 2
                 max_c = cminus + col_off + idx[1] * 2
-        pixmax_mag = count_rate_to_mag(pixmax)
+        # Get the mag equivalent to pixmax.  If pixmax is zero (for a synthetic dark map)
+        # clip lower bound at 1.0 to avoid 'inf' mag and warnings from chandra_aca.transform
+        pixmax_mag = count_rate_to_mag(np.clip(pixmax, 1.0, None))
         pixmags.append(pixmax_mag)
         pix_r.append(max_r)
         pix_c.append(max_c)
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -612,8 +612,8 @@ def get_ax_range(n, extent):
     return a range for the row or col that is divisible by 2 and contains
     at least the requested extent.
 
-    :param n: row or col float value
-    :param extent: half of desired range from n
+    :param n: row or col float value (edge pixel coords)
+    :param extent: half of desired range from n (should include pixel dither)
     :returns: tuple of range as (minus, plus)
     """
     minus = int(np.floor(n - extent))
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -644,9 +644,9 @@ def get_imposter_mags(cand_stars, dark, dither):
     pix_r = []
     pix_c = []
 
-    # Define the 1/2 pixel region as half the 8x8 plus dither
-    row_extent = np.ceil(4 + dither.row)
-    col_extent = np.ceil(4 + dither.col)
+    # Define the 1/2 pixel region as half the 8x8 plus a pad plus dither
+    row_extent = 4 + GUIDE_CHAR.dither_pix_pad + dither.row
+    col_extent = 4 + GUIDE_CHAR.dither_pix_pad + dither.col
     for cand in cand_stars:
         rminus, rplus = get_ax_range(cand['row'], row_extent)
         cminus, cplus = get_ax_range(cand['col'], col_extent)
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -618,9 +618,7 @@ def get_ax_range(n, extent):
     """
     minus = int(np.floor(n - extent))
     plus = int(np.ceil(n + extent))
-    # Use this floor != ceil logic to skip doing anything if the float
-    # n is equivalent to an integer.  This only comes up in test/constructed cases
-    if (np.floor(n) != np.ceil(n)):
+    if (plus - minus) % 2 != 0:
         if n - np.floor(n) > 0.5:
             plus += 1
         else:
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -608,7 +608,8 @@ def check_column_spoilers(cand_stars, ok, stars, n_sigma):
 
 def get_ax_range(n, extent):
     """
-    Given a float pixel row or col value and an integer pixel "extent",
+    Given a float pixel row or col value and an "extent" in float pixels,
+    generally 4 + 1.6 for 8" dither and 4 + 5.0 for 20" dither,
     return a range for the row or col that is divisible by 2 and contains
     at least the requested extent.
 
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -606,22 +606,25 @@ def check_column_spoilers(cand_stars, ok, stars, n_sigma):
     return column_spoiled, rej
 
 
-def get_ax_range(n, extent):
+def get_ax_range(rc, extent):
     """
     Given a float pixel row or col value and an "extent" in float pixels,
     generally 4 + 1.6 for 8" dither and 4 + 5.0 for 20" dither,
     return a range for the row or col that is divisible by 2 and contains
     at least the requested extent.
 
-    :param n: row or col float value (edge pixel coords)
+    :param rc: row or col float value (edge pixel coords)
     :param extent: half of desired range from n (should include pixel dither)
     :returns: tuple of range as (minus, plus)
     """
-    minus = int(np.floor(n - extent))
-    plus = int(np.ceil(n + extent))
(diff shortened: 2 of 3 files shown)


proseco/guide.py
```diff
@@ -635,7 +642,7 @@ def get_ax_range(r, extent):
     # Define the 1/2 pixel region as half the 8x8 plus dither
     row_extent = np.ceil(4 + dither.row)
     col_extent = np.ceil(4 + dither.col)
-    for idx, cand in enumerate(cand_stars):
+    for cand in cand_stars:
```
jeanconn: This is not strictly required for this fix, but idx is redefined below and wasn't used in this context, so that was confusing.

proseco/guide.py
```diff
@@ -606,6 +606,25 @@ def check_column_spoilers(cand_stars, ok, stars, n_sigma):
...
+    return a range for the row or col that is divisible by 2 and contains
+    at least the requested extent
+    :param n: row or col float value
+    :param extent: half of desired range from n
+    :returns: tuple of range as (minus, plus)
+    """
+    minus = int(np.floor(n - extent))
+    plus = int(np.ceil(n + extent))
+    if (np.floor(n) != np.ceil(n)):
+        if n - np.floor(n) > .5:
```
taldcroft: Never write `.5` always `0.5`.

proseco/guide.py
```diff
@@ -606,6 +606,25 @@ def check_column_spoilers(cand_stars, ok, stars, n_sigma):
     return column_spoiled, rej
 
 
+def get_ax_range(n, extent):
+    """
+    Given a float pixel row or col value and an integer pixel "extent",
+    return a range for the row or col that is divisible by 2 and contains
+    at least the requested extent
```
taldcroft: Period at the end and newline.

proseco/guide.py
```diff
@@ -606,6 +606,25 @@ def check_column_spoilers(cand_stars, ok, stars, n_sigma):
...
+    :param n: row or col float value
+    :param extent: half of desired range from n
+    :returns: tuple of range as (minus, plus)
+    """
+    minus = int(np.floor(n - extent))
+    plus = int(np.ceil(n + extent))
+    if (np.floor(n) != np.ceil(n)):
```
taldcroft: I don't understand this logic.  Basically you want `plus` and `minus` to be even, right?  So would this work?
```
# Ensure plus and minus are both even, expanding the range as needed.
if minus % 2 != 0:
    minus -= 1
if plus % 2 != 0:
    plus += 1
```

taldcroft:
This is related to #125 (where I consider imposters to be a type of spoiler).  Unit test with a synthetic star catalog and synthetic dark current is needed.  Probably worth adding a quick unit test of `get_ax_range` as well.

proseco/guide.py
jeanconn: We want (plus - minus) to be even.  If you ensure that both are even (or odd) you do win that, but you can end up adding extra pixels to both ends that really aren't going to be in the 8x8.

jeanconn: With the logic that is in there now, the initial
``` 
minus = int(np.floor(n - extent))
plus = int(np.ceil(n + extent))
```
guarantees you have a odd range (for real float values not exactly at the pixel boundary), and then it adds one more pixel to the side closest to the offset of the star within the center pixel.

taldcroft: If this is about figuring out the readout window then can we use the usual logic for defining a readout window?
```
n_int = int(round(n))  # for `n` in edge pixel coords
plus = n_int + extent
minus = n_int - extent
```

jeanconn: The extent is a float value.  If you want to round it up, this works too, but will get extra pixels.

jeanconn: Where's the usual logic for defining a readout window?  If I should always be taking one of the minus or adding one to the plus anyway, that's fine, but practically, I was trying not to add extra pixels while still making it possible to bin 2x2.  If that's just too ugly we can do something different, but may end up with regression test changes.

taldcroft: Here's one implementation of the "usual logic":
  https://github.com/sot/chandra_aca/blob/master/chandra_aca/aca_image.py#L451
I'm sure you find it in annie as well.

I don't quite follow what you are saying, but the 3 lines I wrote will always give the minimal set of pixels that are centered on `n` and can be binned 2x2.  I suspect it is exactly equivalent to what you wrote.

jeanconn: I'm still confused.  extent is meant to have 4 pixels for the 8x8 image and, generally, plus 1.6 pix for 8x8 dither.  So plus = n_int + 5.6 doesn't quite work, but if I round up in both directions again that probably works.

So, for a star with nominal row=0.7 in edge coordinates, what should be dithered pixel region be in row?

-6 to 6
or
-5 to 7

Because I thought -5 to 7 was the right answer, but I didn't have the usual logic as my reference.

taldcroft:
The docstring said `extent` is an integer, and that does make a difference here.  I was predicating my comments on that assumption.  For non-integer values it might be a different story.

jeanconn:
Oops.  Thanks Tom.  Introduced docstring but fixed.

proseco/tests/test_guide.py
```diff
@@ -286,3 +286,45 @@ def test_guides_include_exclude():
...
+    stars.add_fake_star(row=495.3, col=-502.4, mag=6.0)
+
+    std_info = STD_INFO.copy()
+    std_info.update(n_guide=8)
+    guides = get_guide_catalog(**std_info, stars=stars)
+    assert len(guides) == 8
+
+
+def test_get_ax_range():
+    # Confirm that the ranges from get_ax_range are reasonable for a variety of
```
taldcroft: Use docstrings for tests so they show up in API docs.

proseco/tests/test_guide.py
```diff
@@ -286,3 +286,45 @@ def test_guides_include_exclude():
 
     assert np.all(guides['id'] == [9, 11, 2, 3, 4, 5, 6, 7])
     assert np.allclose(guides['mag'], [10.0, 12.0, 7.1, 7.2, 7.3, 7.4, 7.5, 7.6])
+
+
+def test_edge_star():
+    """
+    Add stars right at row and col max for 8" dither.
```
taldcroft: This test is good but it seems like it does not actually test the bug fix related to get_ax_range.  That would happen with hot pixel imposters right?

Also in both this test and a synthetic hot pixel test, document the exact source of the magic numbers (e.g. 495.3).  I'm never sure whether it is best to have test values like this hardcoded, so it gives a visible regression if characteristics change, or have them computed in the test so tests continue passing.  If you don't care then I'd say the latter.
```
        row_max = CCD['row_max'] - (CCD['row_pad'] + CCD['window_pad'] + r_dith_pad)
        col_max = CCD['col_max'] - (CCD['col_pad'] + CCD['window_pad'] + c_dith_pad)
```

Also in both tests it would be best to use asymmetric dither, esp for imposters since that was the original point of the bug fix.

proseco/guide.py
taldcroft: No problem, not worth a separate PR for tiny cleanup like this.

proseco/tests/test_guide.py
jeanconn: Yeah, I was working on a new test of hot pix imposters Friday but didn't finish.  Will get it in.  And sure, documenting exact source of the numbers sounds good (I at least left a note about nominal dither being required for the test setup, but can also recalculate it directly here.  Figured it was unlikely we'd change CCD characteristics without needing to update all the tests anyway.).  For the edge case test, I don't think it matters if the test is done with asymmetric dither or not, but fine to do so.  For my specific bug also need to test with the "large" dither in each axis.

jeanconn:
I converted "n" to "rc" in get_ax_range, but also didn't like that the use of "ceil" in the way that get_ax_range was being called effectively meant that ACIS observations got a 0.4 arcsec pad, but HRC observations got no padding.  So I added a pad to each and got rid of that use of ceil.  Was that too much @taldcroft  ?

proseco/characteristics_guide.py
```diff
@@ -9,6 +9,12 @@
 # Minimum scaled guide count for thumbs_up
 min_guide_count = 4.0
 
+# Add this padding to region checked for bad pixels (in addition to dither)
+dither_pix_pad = 0.4
+
+# Minimum imposter mag (if dark current is 0 set explicitly to this value)
+min_imposter_mag = 20.0
```
taldcroft: This is actually the *max* value of `imposter_mag`.  Otherwise this and the code that uses it are OK.

jeanconn: Hah.  I should have used a variant on faint.

jeanconn: How about "imposter_mag_for_zero_dark"?

taldcroft: Fine.  Plan B would be to replace `pixmax` with `np.clip(pixmax, 1.0, None)` in the offending line, add a code comment, and be done with it.  I somehow feel like characteristics should be reserved for constants that have a discernible algorithmic impact.  But that is my style and not necessarily correct.  Your choice.  FYI, count rate of 1.0 is very nearly 20th mag.  :smile:.

jeanconn: Sure.  I like that better.  And yes, I saw that count rate of 1 is basically 20th mag.  I considered just setting pixmax to 1.0 before starting the loop over the sums, but didn't want to just add 1.0 in general.  Your clip is better.

jeanconn:
OK, so maybe this is now done?  I didn't see feedback on the 0.4 pad I added to (the HRC) dither boxes.  

I did not update version figuring that could go in after merging (and after any other updates).

Issue #192, opened 2018-12-12
Bright stars may not have large enough mag_err for guide star selection


jeanconn:
Related to #67 

For stars brighter than 6.1 mags, we probably want to set the lower bound on the applied mag_err to 0.1 mags for this check

https://github.com/sot/proseco/blob/52145e0993239c55dd40d75046d9f6c4ce9a8b57/proseco/guide.py#L176

Reviewing the candidate stars, it looks like there are 647 possible candidate stars with MAG_ACA_ERR < 0.1 and MAG_ACA > 5.9 and MAG_ACA < 6.1 (where stars is the whole proseco AGASC).

```
In [63]: worry = stars[(stars['MAG_ACA'] > 5.9) & (stars['MAG_ACA'] < 6.1) & (stars['MAG_ACA_ERR'] < 10) & (stars['CLASS'] == 0) & (stars['ASPQ1'] < 20) 
    ...: & (stars['POS_ERR'] < 3000)]

In [64]: len(worry)
Out[64]: 637
```

taldcroft:
In StarsTable we probably need a `mag_aca_err` value which is explicitly the best estimate of the true systematic catalog error in addition to `mag_err` which is the per-readout error. This is for post-release, nothing is changing right now.

jeanconn:
I think #254 basically closed this issue for me, but @taldcroft comments about needing a new `mag_aca_err` have not been addressed.   Not sure if that needs a new issue.

PR #197, opened 2018-12-17
Guide reporting

jeanconn:


proseco/core.py
```diff
@@ -371,7 +371,7 @@ def __init__(self, data=None, **kwargs):
         # Make printed table look nicer.  This is defined in advance
         # and will be applied the first time the table is represented.
         self._default_formats = {'p_acq': '.3f'}
-        for name in ('yang', 'zang', 'row', 'col', 'mag', 'maxmag', 'mag_err', 'color'):
+        for name in ('yang', 'zang', 'row', 'col', 'mag', 'maxmag', 'mag_err', 'color', 'COLOR1'):
```
jeanconn: I think I want ASPQ1 too

proseco/guide_report.py
```diff
@@ -0,0 +1,290 @@
...
+        y = guide['imp_c']
+        patch = patches.Rectangle((x, y), 2, 2, edgecolor='y', facecolor='none', lw=1.5)
+        ax.add_patch(patch)
+        plt.text(row0 + drow / 2, col0 + dcol - 1, f"box 'mag' {guide['imp_mag']:.1f}",
+                 ha='center', va='center', color='y', fontweight='bold')
+
+    # Plot a box showing the 8x8 image boundaries
+    x = row0 + drow / 2 - 4
+    y = col0 + dcol / 2 - 4
+    patch = patches.Rectangle((x, y), 8, 8, edgecolor='y', facecolor='none', lw=1)
```
jeanconn: I was also trying to decide if there was value in trying to plot some indicator of the edge of the useable fov, though that's probably just applicable for the force-included stars that concerned me earlier.

proseco/guide_report.py
```diff
@@ -0,0 +1,290 @@
...
+    """
+    # NOTE: all coordinates in plot funcs are "edge" coordinates.  Indexing an
+    # ACAImage pixel using `.aca` coordinates implies using the row, col
+    # coordinate of the lower-left corner of that pixel.
+
+    fig = plt.figure(figsize=(8, 4))
+
+    # First plot: star spoilers
+    ax = fig.add_subplot(1, 2, 1)
+    plot_spoilers(guide, cand_guides, ax)
```
jeanconn: Though since this just uses the table of the other candidates to get cand_guides.stars, it would make more sense to just pass cand_guides.stars .

jeanconn:
PS I'm working on the validation notebook, so ping me if you end up wanting to work on that.

proseco/guide_report.py
```diff
@@ -23,7 +23,7 @@
 # Do reporting for at-most MAX_CAND candidates
 MAX_CAND = 50
 COLS = ['id', 'stage', 'forced',
-        'mag', 'mag_err', 'MAG_ACA', 'MAG_ACA_ERR',
+        'mag', 'mag_err',
```
jeanconn: MAG_ACA_ERR is used directly in the spoiler tests.    I thought it was somewhat useful to have it here because I found myself looking it up for the cases where it is not presently reasonable (bright candidates).  But I suppose if the ids are links to the kadi star lookup that's fine.

jeanconn:
PPS Where are the validation notebooks suppose to live at this point?  Not in version control?

taldcroft:
Validation notebooks are in /proj/sot/ska/analysis/proseco/www (which is a link).  I gave up on version control for these, not seeing the value added (but including the `proseco.test(get_version=True)` is key).  

I'm not planning to work on the guide validation notebook.

taldcroft:
On MAG_ACA_ERR, fair enough.  Reverted in f09e3a2.  

The main thing is that in the meantime we put in this cut on mag_err for color=1.5 stars and will probably want something for bright stars, so MAG_ACA_ERR is really not the right thing.  We probably need a `mag_aca_err` value which is explicitly the "correct" systematic catalog error in addition to `mag_err` which is the per-readout error.  This is for post-release, nothing is changing right now.

taldcroft:
And I see this is captured in #192, will update that a bit.

proseco/guide_report.py
taldcroft: Doesn't hurt to have the whole thing and makes the API more consistent.  These may end up as methods anyway in the next round of refactor that I'm planning.

proseco/core.py
taldcroft: Done in 15a0138.

taldcroft:
Merging to just get this into master.  Further improvements or tweaks are likely.

PR #201, opened 2018-12-18
WIP pickle dark cal image params

### LINKED ISSUE #200: Include provenance information in pickle
This includes at least name of the dark cal file if taken from mica or the entire dark map if not.

Maybe do something similar for stars.

taldcroft:
Would close #200, except that it turns out that this opens a can of worms and I'm thinking it isn't worth the trouble.

I went into this trying to develop a re-usable structure that could be applied to `stars` as well and would allow unpickling to just work.  But some problems:

- For a supplied dark map (either ACAImage or ndarray image) this should be pickled as an ACAImage.  Unfortunately it turns out ACAImage doesn't roundtrip through pickle because the `meta` is lost.  It would be feasible to fix ACAImage (requiring a new release of Ska3) or monkey patch the class.  Ugh.
- Even for the nominal case of dark taken from mica, where unpickling would recognize the special tuple in dark, there is an issue that `aca`, `aca.acqs`, `aca.guides`, and `aca.fids` would all find themselves seeing that tuple and running the function to individually get a `dark` value.  Not so nice.

Then I realized that to first order what is being pickled here in the production case isn't *really* getting us that much closer to the actual dark map used in selection.  The kwargs are guaranteed to be the same, and there is still the need for post-processing to add the bad pixels.

So leaving this open for the moment for pondering, but I don't recommend this for 4.3.1.  It still needs testing which isn't so trivial.

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -353,7 +353,7 @@ class ACACatalogTable(Table):
     detector = MetaAttribute()
     sim_offset = MetaAttribute()
     focus_offset = MetaAttribute()
-    dark = MetaAttribute(pickle=False)
+    dark = MetaAttribute()
     stars = MetaAttribute(pickle=False)
     include_ids_acq = IntListMetaAttribute(default=[])
     include_halfws_acq = IntListMetaAttribute(default=[])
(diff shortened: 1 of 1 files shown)


jeanconn:
The most interesting edge case, is when stars are selected, and then the dark map is updated before the loads are approved, right?  So it seems like this can be handled on the starcheck side to some extent.  Worst case we could do the currently-duplicate-code bad pixel check twice (with dark current from a week ago and dark current now) and warn on any really bad hot pixel hits that star selection should be done again.

taldcroft:
Closing this.  If for some reason we end up with a hard driver for this then this PR will still be here for examination.

PR #209, opened 2018-12-21
Updates to support validating proseco pickle contents

taldcroft:
This goes along with validate-pickle-contents.ipynb in the usual place.


taldcroft:
Passing test and not controversial, merging.

PR #216, opened 2018-12-27
WIP: optimize roll

taldcroft:
This has rough draft code for:
- Identify candidate acq and guide stars at different rolls that might improve the catalogs
- Identify discrete roll ranges where those stars are in the FOV

Currently it is more focused on acquisition.  The tricky bit as always is weighing guide vs acq goodness.

Testing with:
```
import numpy as np
from chandra_aca.star_probs import guide_count

aca = get_aca_catalog(18925)
cands = aca.get_better_roll_candidates()
ids_list, ids0 = aca.find_roll_ranges(cands, d_roll=0.25)

ids_list.append(({}, {}, [0, 0]))

acas = []
for d1, d2, roll_offsets in ids_list:
    for roll_offset in (roll_offsets[0], np.mean(roll_offsets), roll_offsets[1]):
        aca_roll = get_aca_catalog(18925, att=[aca.att[0], aca.att[1], aca.att[2] + roll_offset],
                                   dark=aca.dark, raise_exc=True)
        print(f'************* roll_offset = {roll_offset} **************')
        print(f'P2 = {-np.log10(aca_roll.acqs.calc_p_safe()):.2f}')
        n_stars = guide_count(aca_roll.guides['mag'], aca_roll.t_ccd)
        print(f'n_stars = {n_stars:.2f}')
        acas.append(aca_roll)
```
This finds a much better catalog at a roll offset of -19 deg, though sadly that is not legal for this observation.  But it finds a somewhat better catalog in the allowed range.

### DIFF
--- a/proseco/catalog.py
+++ b/proseco/catalog.py
@@ -3,12 +3,14 @@ import numpy as np
 from itertools import count
 
 from astropy.table import Column, Table
+from chandra_aca.star_probs import acq_success_prob
 
 from .core import ACACatalogTable, get_kwargs_from_starcheck_text, MetaAttribute
 from .guide import get_guide_catalog, GuideTable
 from .acq import get_acq_catalog, AcqTable
 from .fid import get_fid_catalog, FidTable
 from . import characteristics_acq as ACQ
+from . import characteristics as ACA
 
 
 def get_aca_catalog(obsid=0, **kwargs):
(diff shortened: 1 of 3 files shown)


proseco/catalog.py
```diff
@@ -481,3 +514,50 @@ def filt(tbl, id_set):
     aca.add_column(idx, index=1)
 
     return aca
+
+ROLL_TABLE = Table.read("""
+pitch   rolldev
+46      0
```
jeanconn: The last table I have from the FOT is included in a thttps://github.com/sot/aca_lts_eval/pull/52/commits/e5bba740c9a91cdba3be2f040d8c32b31122ac6c

taldcroft:
Good point.  My plan A on this was to punt the problem of allowed roll range over to ORviewer (hence the method arg).  ORviewer and the MP's know precisely what roll range they want and can supply it.  The default auto-roll-range code is mostly intended for testing / validation.  But we'll see, things can change...  Here's the flow I'm thinking about right now:

- Select catalogs for all obsids with `get_aca_catalog`.
- Catalogs with P2 < 3.5 (TBR) are always checked for better stars and the results are put into a `better_rolls` list.  ORviewer can see if that list is non-empty and make it known to the user (and potentially available for inspection) via the UI.
- Catalogs with available better rolls can then be optimized.  Again it may be better to put some of the logic on the ORviewer side so that it handles computing the ACA attitudes for the better rolls (taking care of the target roll vs. ACA roll stuff, since proseco doesn't know about target offsets).  Then ORviewer calls `get_aca_catalog` for each of the better attitudes and there is a mechanism for picking the best one based on P2 and guide_count.

taldcroft:
Moved into aca_preview.

PR #218, opened 2018-12-28
More guide test updates

### LINKED ISSUE #203: Improve guide tests
- Docstrings for every test
- Common column synthetic
- Etc.

jeanconn:
More guide test updates including docstring updates and a new synthetic common column test.

Closes #203

### DIFF
--- a/proseco/tests/test_catalog.py
+++ b/proseco/tests/test_catalog.py
@@ -1,5 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import matplotlib
+matplotlib.use('agg')
+
 import pickle
 import pytest
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


taldcroft:
Rebased after #221.

taldcroft:
@jeanconn - added 7f28dec.  This might fall into the category of functional validation and need more investigation.  I'm not sure I understand the results there which are almost independent of the delta mag, and I definitely don't think a 15th mag star (mag0=8 and dmag=7) should ever spoil anything.  That corresponds to 70 e-/sec.

taldcroft:
Proseco AGASC has *every* star in the AGASC that is near a candidate guide or acq star with ASPQ1 > 0.  Here is the MAG_ACA distribution.

image

jeanconn:
Ah.  OK. So probably should have had some useful demonstrable grid of spoiler tests before transitioning to the proseco AGASC, as functionally that means the available spoilers changed during development.  But overall the population of concern is still small.  I think this still falls more into parameter optimization, enhancement, and review as part of #220 .  At worst the current code exclude candidates that are not really spoiled "enough".

PR #221, opened 2018-12-29
Make plot() method behave consistently and correctly

taldcroft:
Plotting was kinda broken with regards to bad stars and the attempt to use just a single method for all classes didn't cut it.  This PR fixes things up by introducing class-specific plot methods that do the right thing.  Along the way I did a lot of refactoring and re-organizing:

- Add a `bad_stars_mask` property
- Add a `get_candidates_mask` method that indicates acceptable candidates (apart from spatial filtering)
- Add a `catalog_type` class attribute to indicate the type of catalog represented by the class.
- Add a helper method `get_catalog_for_plot` (minor catalog munge for plotting)

There was a reason for the madness, namely in development of #216 I was frustrated by the bugs and not being able to evaluate by eye whether a particular candidate should be acceptable at a different roll.

This supercedes #222.

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -425,7 +425,7 @@ class BaseCatalogTable(Table):
         arg of plot_stars where bad stars are plotting in a distinctive color.
 
         """
-        if not hasattr(self, '_bad_stars'):
+        if not hasattr(self, '_bad_stars_mask'):
             self._bad_stars_mask = ~self.get_candidates_mask(self.stars)
         return self._bad_stars_mask
 
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -347,6 +347,9 @@ class BaseCatalogTable(Table):
     - Indexing on 'id' column
 
     """
+    # Catalog type when plotting (None | 'FID' | 'ACQ' | 'GUI')
+    catalog_type = None
+
     def __init__(self, data=None, **kwargs):
         super().__init__(data=data, **kwargs)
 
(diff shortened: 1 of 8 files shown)


taldcroft:
@jeanconn - related to your comment about different definitions of bad_stars, hopefully the new name and code comments help.  I also realized it would make sense to set the `bad_stars_mask` for any stars that are in the `bad_stars_set`.  Will do, though this will only happen for stars that are considered as candidates.

proseco/guide.py
```diff
@@ -288,6 +290,26 @@ def process_include_ids(self, cand_guides, stars):
 
         super().process_include_ids(cand_guides, stars[ok])
 
+    def get_candidates_mask(self, stars):
+        """Get base filter for acceptable candidates.
```
jeanconn: Worth noting that, in general, I don't know what value the label of "acceptable" really has when the CCD-position and more detailed spoiler checks have not been done.  But that's more of an issue for the seeking-the-right-roll PR.

proseco/core.py
```diff
@@ -398,6 +401,62 @@ def get_id_idx(self, id):
...
+        from chandra_aca.plot import plot_stars
+        import matplotlib.pyplot as plt
+
+        kwargs.setdefault('stars', np.array([]))
+
+        # Explicitly set bad_stars so that plot_stars does not apply its
+        # internal version (which is no lot always the right answer).
+        kwargs.setdefault('bad_stars', np.zeros(len(kwargs['stars']), dtype=bool))
+
+        plot_stars(attitude=self.att, ax=ax, **kwargs)
```
jeanconn: Looks good. Do we need more changes/options in chandra_aca.plot_stars?  Other colors etc?

jeanconn:
Thanks, and yeah, I don't know if we should just come up with a new name for the list of stars we are tracking that should be excluded, but the _mask helps some.

proseco/guide.py
taldcroft: "Acceptable" means it *can* be chosen as a candidate.  It doesn't mean it *will* be a candidate.  Conversely, a star that is "not acceptable" will never be chosen as a guide or acq, so this is the initial entry ("base") filter.

proseco/core.py
taldcroft: Maybe, but that's for another day.

proseco/tests/test_guide.py
```diff
@@ -102,14 +102,22 @@ def test_region_contrib():
     assert 426255616 not in selected2['id']
 
 
-@pytest.mark.skipif('not HAS_SC_ARCHIVE', reason='Test requires starcheck archive')
-def test_exclude_bad_star():
-    """
-    Obsid 17896 attitude
-    Will need to find another bad star, as this one is now excluded via VAR
+def test_bad_star_list():
```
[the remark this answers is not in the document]
jeanconn: I think this will make conflicts on #218 which I may let you resolve.

PR #224, opened 2018-12-31
Add n_stars method for catalog fractional count

taldcroft:
This makes the guide fractional count use piecewise linear interpolation instead of mag thresholds for computing the count.  It doesn't really make sense that a 10.29 star counts as 0.5 stars and a 10.31 star is 0.0.  In reality they are functionally equivalent.  

Here is a comparison (taken from a notebook referenced in the 2019-01-09 SSAWG notes).  Red is the new curve, magenta is the old one.  The new implementation is exactly the same at the reference values, but on average more optimistic.  From analysis in the notebook this is probably justified but needs discussion.

image

Note that I'm using this now in #216 as the thresholding obscures changes in the guide catalog.

### DIFF
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -105,12 +105,11 @@ class GuideTable(ACACatalogTable):
         SSAWG.  The implementation here does a piecewise linear interpolation
         between the reference mag - fractional count points instead of the
         original "threshold interpolation" (nearest neighbor mag <= reference
-        mag).
+        mag).  Approved at 16-Jan-2019 SSAWG.
 
         :returns: fractional count
 
         """
-        # The bright limit does not scale.
         t_ccd = self.t_ccd
         mag1 = snr_mag_for_t_ccd(t_ccd, ref_mag=10.0, ref_t_ccd=-10.9)
         mag2 = snr_mag_for_t_ccd(t_ccd, ref_mag=10.2, ref_t_ccd=-10.9)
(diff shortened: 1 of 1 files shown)


jeanconn:
Are you putting this in proseco so you don't need to cut a version of chandra_aca to use it?

taldcroft:
We cannot put it into flight chandra_aca without SSAWG discussion.  So if/when this migrates out to chandra_aca we can change the `n_stars` method to just call that.  I suppose we could make a new version of chandra_aca and put it into ska3-test (if / when that exists).

jeanconn:
Gotcha.  I figured we could add methods to chandra_aca without SSAWG discussion, just its use would eventually require discussion.  I'm assuming that there would be no huge advantage to using it for current star catalog evaluation.

taldcroft:
I would plan to change `star_probs.guide_count()` to make this the official one-and-only-one metric that gets used everywhere.  So that's where the "we can't promote to flight before SSAWG approves" came from.

proseco/guide.py
```diff
@@ -95,6 +96,36 @@ def reject(self, reject):
...
+        suggested at the 7-Mar-2018 SSAWG and agreed upon at the 21-Mar-2018
+        SSAWG.  The implementation here does a piecewise linear interpolation
+        between the reference mag - fractional count points instead of the
+        original "threshold interpolation" (nearest neighbor mag <= reference
+        mag).
+
+        :returns: fractional count
+
+        """
+        # The bright limit does not scale.
```
jeanconn: I think this comment doesn't apply here.

taldcroft:
Replaced by chandra_aca 4.25.

PR #227, opened 2019-01-05
Make AcqProbs ref to parent acqs be a weakref

taldcroft:
Basically the same as #226, but applied to the 4.3.x branch.

The refactoring in master made it not possible to cherry pick the commits so I just applied the diffs from #226 by hand (in a "curated" way to make it right for 4.3.x).

Memory leak fix.  The plot below shows memory usage (blue) for selecting 300 full catalogs on Mac:
plot

Testing:
- Pass unit tests on Mac and HEAD linux-64 (ska3/flight), GRETA linux-32 (ska3/matlab32/test).
- Write acq report for obsid 21068 and confirm no unexpected diffs in `index.html` from 4.3.1.

### DIFF
--- a/proseco/acq.py
+++ b/proseco/acq.py
@@ -6,6 +6,8 @@ Get a catalog of acquisition stars using the algorithm described in
 https://docs.google.com/presentation/d/1VtFKAW9he2vWIQAnb6unpK4u1bVAVziIdX9TnqRS3a8
 """
 
+import weakref
+
 import numpy as np
 from scipy import ndimage, stats
 from scipy.interpolate import interp1d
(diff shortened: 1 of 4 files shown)


proseco/acq.py
```diff
@@ -126,6 +128,23 @@ class AcqTable(ACACatalogTable):
...
+        super().__setstate__(state)
+
+        # This could be a cand_acqs table or acqs table, so check if
+        # ``cand_acqs`` has something, and if so then create the weakref in
+        # each of the AcqProbs objects stored in the ``probs`` column.  TO DO:
+        # make two separate classes AcqTable and CandAcqTable to avoid this
+        # contextual hack.
+        if self.cand_acqs is not None:
+            for probs in self.cand_acqs['probs']:
+                probs.acqs = weakref.ref(self)
```
jeanconn: This looks reasonable to me, but I'm not as clear about how we get to this point in the code.  With that in mind does this need any new tests besides the two assert statements added in test_catalog?

taldcroft: This is basically the unpickling post-processing code, so gets called when an AcqTable is unpickled.  The existing tests (even without the two new asserts) will give an exception without this code.  I've added a comment in the test code to explain why the asserts there are "strong" tests of this new functionality, agreed it wasn't obvious just from the asserts.

mbaski:
These updates work on Windows and appear to resolve the memory leak (memory doesn't keep increasing each time a DOT is generated from ORViewer).  The unit tests pass "105 passed in 55.53 seconds" on Windows and the DOT and associated files generated with this updated version of proseco using the dbtest/11706/DEC2418_11706.mat workspace are exactly the same as before the update (except for timestamps, of course).

PR #238, opened 2019-01-14
Add star 350392600 (3.6 mag in obsid 21252 cat mag 7.6) to bad star list

jeanconn:
Add star 350392600 (3.6 mag in obsid 21252 cat mag 7.6) to bad star list


jeanconn:
Not sure if we'll have a chance to update these characteristics before we move to more dynamic fetching of this kind of content anyway.

PR #240, opened 2019-01-18
Add manual bad star 1158290496 (bad position / PM)

taldcroft:
See https://github.com/sot/starcheck/pull/290


taldcroft:
I tested with obsid 19763.  Current master selects 1158290496 as BOT, while this branch selects for neither guide nor acq.

jeanconn:
Also, it looks to me like this star is known to have [high proper motion](http://simbad.u-strasbg.fr/simbad/sim-coo?Coord=348.95657694+-58.08256833&CooFrame=FK5&CooEpoch=2000&CooEqui=2000&CooDefinedFrames=none&Radius=2&Radius.unit=arcsec&submit=submit+query&CoordList=)

It just doesn't have correct values in AGASC 1.7 .  Given the risk associated with this kind of star we should probably take an action to review and update positions and proper motions in the catalog.

PR #241, opened 2019-01-19
Prelim products (proseco-only) review

taldcroft:
@jeanconn : this is a super-simple initial cut which at least makes a plot of every observation.  I was using in ipython and then looking at them in Mac preview.
```
run review
acas = get_acas('jan2819.pkl')
make_starcat_plots(acas, 'jan2819')
```

For JAN2819 prelim, I didn't see any obvious constellation problems, but several ERs have stars that are probably too close to the edge.  Should be straightforward to replicate the starcheck edge checks here, print the star ids and supply those ids to James for `exclude_ids_guide`.


proseco/preview.py
```diff
@@ -0,0 +1,317 @@
...
+        for entry in self:
+            self.check_position_on_ccd(entry)
+
+        self.check_guide_geometry()
+        self.check_acq_p2()
+
+    def check_guide_geometry(self):
+        """Check for guide stars too tightly clustered
+
+        (1) Check for any set of n_guide-2 stars within 500" of each other.
```
jeanconn: The n-2 check seems a fine place to start.  Not sure how to figure out a reasonable tolerance but 500" doesn't seem too crazy.

jeanconn:
For ERs, I'd also like a check that there are 3 brighter than 9, which I can add.

jeanconn:
Also @taldcroft, overall this prelim review tool looks great and is instantly helpful.  Thanks for working on it this weekend.

taldcroft:
Superceded by `aca_preview` (https://github.com/sot/aca_preview/pull/1)

PR #245, opened 2019-01-24
Add 3 pixels to edge margin for guide star selection (4.3.x patch)

jeanconn:
Add 3 pixels to edge margin for guide star selection (4.3.x patch)

This PR makes a new characteristics value proseco.characteristics_guide.CCD['guide_extra_pad'] that is set to 3 pixels.  This is used in guide star selection as part of the initial filter on stars so these stars that do not have at least these 3 extra pixels margin from the dither padded + window padded + (row/col) padded edge will not make the initial cut of candidates.


jeanconn:
The only thing driving it is a copy of the test that was already there.

PR #246, opened 2019-01-25
Set higher threshold for thumbs_up for ER guide stars (3 bright, 8 of 8 requested)

jeanconn:
Set higher threshold for ER guide stars (3 bright, 8 of 8 requested)

### DIFF
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -100,8 +100,19 @@ class GuideTable(ACACatalogTable):
             out = 1
         elif len(self) == 0:
             out = 0
+        # If this is an ER as inferred by number of requested guide stars, apply
+        # more requirements on the guide catalog to get a thumbs_up = 1.
+        elif self.n_guide > 6:
+            n_bright = np.count_nonzero(self['mag'] < GUIDE_CHAR.er_bright_thresh)
+            count = guide_count(self['mag'], self.t_ccd)
+            # Confirm it has all of the requested number of guide stars, has
+            # at least n stars brighter than the er_bright_thresh, and has
+            # the minimum guide count.
+            out = int((len(self) == self.n_guide) &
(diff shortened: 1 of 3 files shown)


proseco/tests/test_catalog.py
```diff
@@ -248,6 +248,25 @@ def test_bad_obsid():
     assert aca.fids.thumbs_up == 0
 
 
+def test_check_guide_thumbs_for_ers():
+    dark = DARK40.copy()
+    stars1 = StarsTable.empty()
+    stars1.add_fake_constellation(mag=[8.0, 8.0, 8.0, 10.0, 10.0, 10.0, 10.0, 10.0],
```
jeanconn: Should update this to use the thresholds in characteristics if we decide to go with something like this change.

proseco/guide.py
```diff
@@ -100,8 +100,19 @@ def thumbs_up(self):
             out = 1
         elif len(self) == 0:
             out = 0
+        # If this is an ER as inferred by number of requested guide stars, apply
+        # more requirements on the guide catalog to get a thumbs_up = 1.
+        elif self.n_guide > 6:
```
jeanconn: It looks like I don't have a trivial way to impliment the higher P2 lim on thumbs_up in acq.py because I have no way of telling ERs from ORs.  For guide I think we can infer from the requested number of guide stars.

jeanconn:
This one I thought might have some value for reducing back-n-forths, but I'm guessing from the lack of feedback that you don't want to do this one.

jeanconn:
I suppose inferring an ER from n_guide isn't awesome, but also not sure if the lack of feedback just means @taldcroft hasn't seen this in too much github traffic.

taldcroft:
Closed in favor of https://github.com/sot/aca_preview/issues/37.

PR #248, opened 2019-01-27
Add manual bad star 1158290496 (4.3.x backport)

taldcroft:
Backport #240 to 4.3.x.


jeanconn:
I also don't really see the huge driver to add another single star to the bad star list when this will be caught downstream, but it also isn't a big change.

taldcroft:
This is a habit from astropy.  By default you always backport bug fixes, and I consider this a "bug".  Doesn't matter if downstream tools catch this before going to the spacecraft.

PR #255, opened 2019-01-30
Fix missing f in f-string

taldcroft:
Merging now, passes tests in ska3-flight.


taldcroft:
The backport to 4.3.x is done along with a version bump to 4.3.4.

PR #258, opened 2019-02-06
Try new candidate combinations for clustered guide stars

jeanconn:
This PR takes steps to try different candidate star combinations to attempt to satisfy 3 checks that the guide stars are not too "clustered".

### DIFF
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -196,7 +196,7 @@ class GuideTable(ACACatalogTable):
             n_pass, n_tests = run_select_checks(cands)  # This function knows how many tests get run
             if n_pass == n_tests:
                 self.log(f'Selected stars passing all tests at combination {n_tries}',
-                         tried_combinations=n_tries > 1)
+                         tried_combinations=n_tries)
                 return cands
             else:
                 if n_pass not in select_results:
@@ -207,7 +207,7 @@ class GuideTable(ACACatalogTable):
         max_n_pass = max(select_results)
         self.log(
             f'Settled for combination that satisfied {max_n_pass} cluster checks',
-            warning=True, tried_combinations=True)
+            warning=True, tried_combinations=n_tries)
         return select_results[max_n_pass]
 
     def search_stage(self, stage):
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -177,7 +177,7 @@ class GuideTable(ACACatalogTable):
 
         def index_combinations(n, m):
             seen = set()
-            for n_tmp in range(m + 1, n + 1):
+            for n_tmp in range(m, n + 1):
                 for comb in combinations(range(n_tmp), m):
                     if comb not in seen:
                         seen.add(comb)
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -2,6 +2,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 import numpy as np
+from itertools import combinations
 
 import chandra_aca.aca_image
 from chandra_aca.transform import mag_to_count_rate, count_rate_to_mag
@@ -14,6 +15,8 @@ from . import characteristics_guide as GUIDE
 from .core import bin2x2, ACACatalogTable, MetaAttribute, AliasAttribute
 
 CCD = ACA.CCD
+
+STAR_PAIR_DIST_CACHE = {}
 
 
 def get_guide_catalog(obsid=0, **kwargs):
@@ -39,6 +42,8 @@ def get_guide_catalog(obsid=0, **kwargs):
 
     :returns: GuideTable of acquisition stars
     """
+    global STAR_PAIR_DIST_CACHE
+    STAR_PAIR_DIST_CACHE.clear()
 
     guides = GuideTable()
     guides.set_attrs_from_kwargs(obsid=obsid, **kwargs)
@@ -82,6 +87,7 @@ class GuideTable(ACACatalogTable):
 
     cand_guides = MetaAttribute(is_kwarg=False)
     reject_info = MetaAttribute(default=[], is_kwarg=False)
+    required_optimization = MetaAttribute(default=False, is_kwarg=False)
 
     def reject(self, reject):
         """
@@ -142,8 +148,10 @@ class GuideTable(ACACatalogTable):
         for idx, stage in enumerate(GUIDE.stages, 1):
             already_selected = np.count_nonzero(cand_guides['stage'] != -1)
 
-            # If we don't have enough stage-selected candidates, keep going
-            if already_selected < n_guide:
+            # If we don't have enough stage-selected candidates, keep going.
+            # Enough is defined as the requested n_guide + a "surplus" value
+            # in characteristics
+            if already_selected < (n_guide + GUIDE.surplus_stars):
                 stage_ok = self.search_stage(stage)
                 sel = cand_guides['stage'] == -1
                 cand_guides['stage'][stage_ok & sel] = idx
@@ -153,13 +161,49 @@ class GuideTable(ACACatalogTable):
                 self.log(f'Quitting after stage {idx - 1} with {already_selected} stars', level=1)
                 break
         self.log('Done with search stages')
-        selected = cand_guides[cand_guides['stage'] != -1]
-        selected.sort(['stage', 'mag'])
-        if len(selected) >= self.n_guide:
-            return selected[0:n_guide]
+        stage_cands = cand_guides[cand_guides['stage'] != -1]
+        stage_cands.sort(['stage', 'mag'])
+        selected = self.select_catalog(stage_cands[0:n_guide + GUIDE.surplus_stars])
         if len(selected) < self.n_guide:
             self.log(f'Could not find {self.n_guide} candidates after all search stages')
-            return selected
+        return selected
+
+    def select_catalog(self, stage_cands):
+        """
+        For the candidates selected at any stage (stage-assigned candidates?) select the
+        first combination that satisfies the most number of extra tests.
+        Here the extra tests are just the 3 checks for guide stars that are
+        too tightly clustered.
+        """
+        self.log(f'Selecting catalog from {len(stage_cands)} stage-selected stars')
+        # Set some arrays to save "check status" and combination identifiers so
+        # we only have to search through the candidate combinations ones.  The plan is
+        # if no combination satisfies all the checks, lower the standard (required success
+        # on n checks) and repeat, but do this just searching through the previously
+        # determined status.
+        saved_status = []
+        saved_idxs = []
+
+        # I should come back to this and see if the "min" is needed or if the combinations
+        # code runs fine even if we have fewer-than-expected stage_cands to start
+        choose_n = min(len(stage_cands), self.n_guide)
+
+        for comb in combinations(range(len(stage_cands)), choose_n):
+            cluster_status = run_comb_checks(stage_cands[[comb]])
+            saved_idxs.append(comb)
+            saved_status.append(cluster_status)
+            if len(saved_idxs) > 1:
+                self.required_optimization = True
+            if np.all(cluster_status):
+                self.log(f'Selected stars after trying {len(saved_idxs)} combinations')
+                return stage_cands[[comb]]
+        for n_success in [2, 1, 0]:
+            for idxs, status in zip(saved_idxs, saved_status):
+                if np.sum(status) == n_success:
+                    self.log(
+                        f'Settled for combination that satisfied {n_success} of 3 cluster checks',
+                        warning=True)
+                    return stage_cands[[idxs]]
 
     def search_stage(self, stage):
         """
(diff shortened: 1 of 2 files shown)


proseco/guide.py
```diff
@@ -153,13 +160,49 @@ def run_search_stages(self):
...
+        for comb in combinations(range(len(stage_cands)), choose_n):
+            cluster_status = run_comb_checks(stage_cands[[comb]])
+            saved_idxs.append(comb)
+            saved_status.append(cluster_status)
+            if len(saved_idxs) > 1:
+                self.required_optimization = True
+            if np.all(cluster_status):
+                self.log(f'Selected stars after trying {len(saved_idxs)} combinations')
+                return stage_cands[[comb]]
+        for n_success in [2, 1, 0]:
```
jeanconn: For a first go, instead of deciding which of the checks was most important or most likely to fail, I just set the code to allow for 2 of 3, 1 of 3, and then no success when running through the checks.

proseco/guide.py
```diff
@@ -153,13 +160,49 @@ def run_search_stages(self):
...
+            for idxs, status in zip(saved_idxs, saved_status):
+                if np.sum(status) == n_success:
+                    self.log(
+                        f'Settled for combination that satisfied n_success of 3 cluster checks',
```
jeanconn: Oops, missing format part of string.

jeanconn:
I've got a set of 400 for which the code works.  "Fixed" was probably to strong for a statement for they resulting catalog; what I'll really need to do is run all the optimized catalogs through aca_preview and see what else the optimization might break.

proseco/guide.py
```diff
@@ -153,13 +161,49 @@ def run_search_stages(self):
...
+                        warning=True)
+                return stage_cands[[idxs]]
```
jeanconn: And I think this is the wrong indent.

jeanconn:
@taldcroft mentioned that the combination checking there might be value in trying the first "next" candidate as a possible replacement for more of the earlier stars sooner in the sort.  My understandings is right now we've got

a)

0 1 2 3 4
0 1 2 3 5
0 1 2 3 6
0 1 2 3 7

And there could be value to

b)

0 1 2 3 4
0 1 2 3 5
0 1 2 4 5
0 1 3 4 5

I think I'm leaning a bit more toward the plain combinations, because I think we don't want to hurry to get to replacing the "best" star with a star that didn't even make the first cut?  But a hard problem.  If our stage "labels" were more meaningful, if would be easier to at least optimize a bit by that (starting with the candidates that all share the "next candidate" stage only).

jeanconn:
Also worth brainstorming in this thread what debugging information is going to be most helpful to the ACA reviewer in the guide report for a tricky catalog that has already been "optimized" by this code.  I was shopping just adding an attribute to the catalog to track if the "not first choice" catalog was selected from the available candidates due to this optimization, and added some log messages.  

Maybe something like added text "(not selected due to overall catalog issues (clusters))" as well.

jeanconn:
And it looks like 2 observations (16147 and 19969) since 2006:001 get through optimization/fixes here with a catalog that *still* doesn't satisfy all checks.  16147 has only 6 candidates to start and fixes don't help enough.

And it looks like 19969 is due a bug most likely in the starcheck database but will review it in a proseco context as well.

```
In [21]: cat = get_aca_catalog(19969)
Obsid 19969 didn't satisfy the cluster checks after optimization
In [22]: cat.guides.n_guide
Out[22]: 2
```

proseco/guide.py
```diff
@@ -39,6 +42,8 @@ def get_guide_catalog(obsid=0, **kwargs):
 
     :returns: GuideTable of acquisition stars
     """
+    global STAR_PAIR_DIST_CACHE
```
jeanconn: @taldcroft I'm not sure about the most reasonable place to clear the cache of star-pair distances.  I think this code would only be called via get_guide catalog.

jeanconn:
For this PR, I think this combination selection works, but could use a hand making sure that the cache clearing is reasonable (and style bits like having reasonable variable names and useful docs/comments).

Then it needs tests.

For assessing possible improvements to the order in which we go through combinations, @taldcroft were you thinking something like reviewing guide_count of the first match for each combination-selection-algorithm (to vet the algorithm, not as part of the algorithm?

taldcroft:
I think we should prefer 1 2 3 4 5 over 0 1 2 3 15.  The rationale is that for the issue of a bad dr95, a *single bad star* is typically the culprit. Since the "5th" best star is on average brighter and has fewer problems than the "16th" best, the 1 2 3 4 5 catalog is, statistically, lower risk.

Another way to think about it: if you have 16 candidate stars (all nicely distributed).  So 0 1 2 3 4 is the starting catalog.  But then you find that star 0 must be excluded for some reason, so what is your next choice for the catalog?  It would be 1 2 3 4 5, not 1 2 3 4 15.

And here is the code which is efficient and basically sorts the combinations by max(index).
```
from itertools import combinations

def index_combinations(n, m):
    seen = set()
    for n_tmp in range(m + 1, n + 1):
        for comb in combinations(range(n_tmp), m):
            if comb not in seen:
                seen.add(comb)
                yield comb

for comb in index_combinations(6, 3):
    print(comb, '  ', max(comb))
(0, 1, 2)    2
(0, 1, 3)    3
(0, 2, 3)    3
(1, 2, 3)    3
(0, 1, 4)    4
(0, 2, 4)    4
(0, 3, 4)    4
(1, 2, 4)    4
(1, 3, 4)    4
(2, 3, 4)    4
(0, 1, 5)    5
(0, 2, 5)    5
(0, 3, 5)    5
(0, 4, 5)    5
(1, 2, 5)    5
(1, 3, 5)    5
(1, 4, 5)    5
(2, 3, 5)    5
(2, 4, 5)    5
(3, 4, 5)    5

for comb in combinations(range(6), 3):
    print(comb, '  ', max(comb))
(0, 1, 2)    2
(0, 1, 3)    3
(0, 1, 4)    4
(0, 1, 5)    5
(0, 2, 3)    3
(0, 2, 4)    4
(0, 2, 5)    5
(0, 3, 4)    4
(0, 3, 5)    5
(0, 4, 5)    5
(1, 2, 3)    3
(1, 2, 4)    4
(1, 2, 5)    5
(1, 3, 4)    4
(1, 3, 5)    5
(1, 4, 5)    5
(2, 3, 4)    4
(2, 3, 5)    5
(2, 4, 5)    5
(3, 4, 5)    5
```

taldcroft:
> I was shopping just adding an attribute to the catalog to track if the "not first choice" catalog was selected from the available candidates due to this optimization, and added some log messages.

I'd rather not add a specific attribute for this, instead just do this with logging messages.  I'm not sure what you can do with that True/False attribute in the report besides emit some text saying that the spatial optimization was required.  So instead just add that as a log message.

Remember that log messages can have arbitrary keys of your choosing, so you can add a message with a key `required_optimization = True` that the guide reporting code can respond to.

proseco/guide.py
```diff
@@ -153,13 +161,49 @@ def run_search_stages(self):
...
+        # determined status.
+        saved_status = []
+        saved_idxs = []
+
+        # I should come back to this and see if the "min" is needed or if the combinations
+        # code runs fine even if we have fewer-than-expected stage_cands to start
+        choose_n = min(len(stage_cands), self.n_guide)
+
+        for comb in combinations(range(len(stage_cands)), choose_n):
+            cluster_status = run_comb_checks(stage_cands[[comb]])
```
taldcroft: This code is advertised above to be extensible for other checks, and is named in a general way, but the surrounding code here is hardwired to just the cluster checks.  (Everything will break if you add a 4th test).  I think it should be more like:
```
select_results = {}
for comb in index_combinations(len(stage_cands), choose_m):   # Usually use "m", as in N choose M.
    cands = stage_cands[list(comb)]  # (note that [(1,2)] is not the same as list((1,2))
    n_pass, n_tests = run_select_checks(cands)  # This function knows how many tests get run
    if n_pass == n_tests:
        return cands
    else:
        if n_pass not in select_results:
            # Keep a copy of the first cands table with n_pass passing tests
            select_results[n_pass] = cands

if len(select_results) == 0:
    # Fall back to the original
else:
    # Return the catalog with the maximum n_pass key value
    return select_results[max(select_results)]
```

proseco/guide.py
```diff
@@ -153,13 +161,49 @@ def run_search_stages(self):
...
+            saved_idxs.append(comb)
+            saved_status.append(cluster_status)
+            if len(saved_idxs) > 1:
+                self.required_optimization = True
+            if np.all(cluster_status):
```
taldcroft: Will be OBE, but in general you should use the Python built-in `all` (`any`, `sum`) when dealing with small Python lists.  This will be substantially faster than the numpy version.  Not that it matters, but it is also a style thing because using numpy methods implies the thing is a numpy array.

proseco/guide.py
```diff
@@ -814,3 +858,55 @@ def spoiled_by_bad_pixel(cand_guides, dither):
...
+
+def check_single_cluster(cand_guide_set, threshold, n_minus):
+    """
+    Confirm that a set of stars satisfies the minimum separation threshold when n_minus
+    stars are removed from the set.  For example, for a threshold of 1000, n_minus of 1,
+    and an input set of candidates stars with 5 candidates, confirm that, for any 4 of
+    the 5 stars (5 minus n_minus = 1), one pair of those stars is separated by at
+    least 1000 arcsecs.
+
+    :rtype: bool
```
taldcroft: Docs.

proseco/guide.py
taldcroft: Here is fine.  Not sure why you are declaring the var to be global here, that is not required for modifying this dict.

Plan B for more "correctness" (at the cost of a little "messy") would be creating a `dist_cache` var in `select_catalog` and then passing it down through the layers.  I could go either way.

jeanconn:
"I think we should prefer 1 2 3 4 5 over 0 1 2 3 15. The rationale is that for the issue of a bad dr95, a _single bad star_ is typically the culprit."  Good point!  Thanks!  As much part of me probably wants the best and the brightest when I look at it later in the ground solution, the single bad star is a more credible issue onboard.

I suppose my order of combinations would be a little safer if it used the stage information better (limiting the allowed candidates in chunks by stage), but your sort seems fine as well.  Will validate.

taldcroft:
> ... if it used the stage information better (limiting the allowed candidates in chunks by stage), but your sort seems fine as well

The ordering I proposed does this implicitly since the stage candidates table is already sorted by stage.

proseco/guide.py
taldcroft: For now, just removing the global is good enough.

jeanconn:
Yes, I recognized that.  

Was commenting this more on it as a defect in my ordering and recognizing a possible middle-ground technique that could be applied to my ordering without going quite as quickly to a likelihood of swapping out the best star.  But not really better in principle than what you suggest if as-you-say we just want to not have that one bad star that perturbs dr95.

proseco/guide.py
jeanconn: If you can't satisfy all 3 tests, you still want the first combination (in your preferred order) that would satisfy 2 of 3.  How is the "first combination in your preferred order" part guaranteed with the use of the select_results dict?

jeanconn: Oh wait, I get it, you win that as part of "if n_pass not in select_results".

jeanconn: Thought I'm still not sure how we'd ever get to "if len(select_results) == 0"? n_pass == 0 is a valid key.

taldcroft: Oops that's right, you always get an n_pass.  So it's even simpler!  Just `return select_results[max(select_results)]` at the end without the conditional block.

jeanconn: :+1:   

Though I think for belt-and-suspenders on weird cases, the modified run_select_checks should be ready for the case where no tests could be run (I think this is just the only 3-or-fewer-candidates-in-the-field case) and should just return 0, 0 in that case instead of unhandled exception.

taldcroft: Well `n_tests` should still be whatever it is.  There is no a priori requirement that if one case "cannot run tests", that this will apply to subsequent combinations.  The only thing you lose by letting it continue is a tiny performance hit.  And in fact you might want to return -1 for n_pass (no tests were able to run), which is worse than 0 (no tests passed).

jeanconn: Yep, I just wasn't sure if I had to handle it differently, but the code that is there (with the list of statuses) will basically work.  I think I'll leave the -1 as an idea for RFE.

proseco/guide.py
jeanconn: Not sure if I address the Docs issue here.

proseco/guide.py
```diff
@@ -153,13 +159,56 @@ def run_search_stages(self):
...
+        # I should come back to this and see if the "min" is needed or if the combinations
+        # code runs fine even if we have fewer-than-expected stage_cands to start
+        choose_m = min(len(stage_cands), self.n_guide)
+
+        for n_tries, comb in enumerate(index_combinations(len(stage_cands), choose_m), start=1):
+            cands = stage_cands[list(comb)]  # (note that [(1,2)] is not the same as list((1,2))
+            n_pass, n_tests = run_select_checks(cands)  # This function knows how many tests get run
+            if n_pass == n_tests:
+                self.log(f'Selected stars passing all tests at combination {n_tries}',
+                         tried_combinations=n_tries > 1)
```
jeanconn: Is the enumerate use here and the new/arbitrary tried_combinations key somewhat sensible to you?

taldcroft: Why not just put in `n_tries`.  Then downstream code can do what it wants with that, including testing for `n_tries > 1` or `n_tries > 3` or whatever.

taldcroft: The enumerate is good.

jeanconn: OK, I went back and forth on just n_tries as well because I wasn't sure what the value should really be for the block at the bottom when n_pass != n_tests .  But I think the last value is reasonable.

jeanconn: Also trying to make sure that there's no case where 

index_combinations(len(stage_cands), choose_m)

would error or have no combinations.

jeanconn: Ah, so yeah, for index_combinations I'm in a little trouble in small sets of candidates, where I expected N choose M to just give me N if M == N .

In [12]: for comb in combinations([1, 2, 3, 4], 4):
    ...:     print(comb)
    ...:     
(1, 2, 3, 4)

In [13]: for comb in index_combinations(4, 4):
    ...:     print(comb)
    ...:

proseco/guide.py
```diff
@@ -153,13 +159,58 @@ def run_search_stages(self):
...
+        For the candidates selected at any stage (stage-assigned candidates?) select the
+        first combination that satisfies the most number of extra tests.
+        Here the extra tests are just the 3 checks for guide stars that are
+        too tightly clustered.
+        """
+        self.log(f'Selecting catalog from {len(stage_cands)} stage-selected stars')
+
+        def index_combinations(n, m):
+            if n == m:
+                yield range(n)
```
jeanconn: Sacrilege?

taldcroft: Yes.  That was my bug, it should be `range(m, n + 1)` not `range(m + , n + 1)`.

jeanconn: Ah.  That makes more sense.  Thanks!

jeanconn:
I reran the validation at https://nbviewer.jupyter.org/url/cxc.harvard.edu/mta/ASPECT/proseco/validation-4.4/small_box_PR_validation.ipynb and I think this is good and reasonable.  I was also working on some more functional tests, but those could go in a separate PR if we just want to get this merged.

taldcroft:
:tada:

jeanconn:
Thanks for the assist with the indexed_combinations on this!

PR #269, opened 2019-02-13
Fix _baserepr_ so it doesn't deepcopy all meta

taldcroft:
It turns out the base code to just print catalog tables nicely was doing some heavy lifting behind the scene.  Namely making a full copy of `meta`, which potentially includes star catalogs and dark cals.  This fixes that and is *necessary* for https://github.com/sot/sparkles/pull/48.

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -822,7 +822,8 @@ class ACACatalogTable(BaseCatalogTable):
             if name in names:
                 self[name].format = self._default_formats.pop(name)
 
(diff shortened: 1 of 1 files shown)


taldcroft:
This passes tests relevant tests.  Unrelatedly however chandra_aca 4.25 breaks one proseco test and that needs to be addressed.

taldcroft:
I'm merging to @jskrist can use.

PR #270, opened 2019-02-14
Fix tests related to guide_count

taldcroft:
Requires https://github.com/sot/chandra_aca/pull/72

### DIFF
--- a/proseco/tests/test_catalog.py
+++ b/proseco/tests/test_catalog.py
@@ -189,7 +189,7 @@ def test_pickle():
     """
     stars = StarsTable.empty()
     stars.add_fake_constellation(mag=10.0, n_stars=5)
-    aca = get_aca_catalog(stars=stars, raise_exc=True, **STD_INFO)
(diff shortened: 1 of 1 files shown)


proseco/tests/test_catalog.py
```diff
@@ -189,7 +189,7 @@ def test_pickle():
     """
     stars = StarsTable.empty()
     stars.add_fake_constellation(mag=10.0, n_stars=5)
-    aca = get_aca_catalog(stars=stars, raise_exc=True, **STD_INFO)
+    aca = get_aca_catalog(stars=stars, dark=DARK40, raise_exc=True, **STD_INFO)
```
taldcroft: This was always only selecting 4 stars and just squeaking by with guide count of 4.0.  That was not the intent of this test

jeanconn: Gotcha.  And yes, those 10.0 mag stars can be tricky on an average dark map.  Looks like a reasonable fix.

taldcroft: For the record can you punch the approve?

PR #272, opened 2019-02-14
Read AGASC supplement and add to bad_stars_set if available

### LINKED ISSUE #261: Add capability to use regularly updated bad star list and history file
Could we agree on a spec for https://github.com/sot/ska-projects/issues/92 and set proseco and aca_preview to use it all in proseco 4.4?  I'm also talking about having Mark add it to the list of files that would be synced.

We could finish process for updating the file and such after the release.

This would be a stretch for 4.4.

taldcroft:
Closes #261.

Will open another issue to capture the idea of getting mags from a history file, for now descoping 261 to just be bad stars. To do:

- [x] Tests

### DIFF
--- a/proseco/characteristics.py
+++ b/proseco/characteristics.py
@@ -1,3 +1,8 @@
+import os
+import warnings
+from pathlib import Path
+from astropy.table import Table
+
 CCD = {'row_min': -512.0,
        'row_max': 512.0,
        'col_min': -512.0,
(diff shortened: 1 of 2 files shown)


taldcroft:
I was on the edge about warning, but tests do fail without it.   I just wanted to be sure that proseco would not fail to even import if the file is missing.

PR #273, opened 2019-02-15
Reduce allowed position error from 3.0 to 1.25 arcsec

### LINKED ISSUE #242: Consider changing guide POS_ERR threshold
The current POS_ERR threshold for guide selection allows errors up to 3.0 arcsec.  This is actually quite large and we likely should not use such stars.  Fortunately there are rather few candidate guide stars with position error above 1.0 arcsec, about 0.2% (about 800 of 432k), so 1.0 arcsec might be a natural more conservative limit.

image
The plot above is based on the miniagasc_1p7, filtered by `GuideTable.get_candidates_mask()`.

In the meantime we should put in warnings to `aca_preview`, maybe warning at 1" and critical at 2".

taldcroft:
This cuts about 90 out of 432000 stars that were allowed in with the 3.0 arcsec limit.  Systematic positional offsets can lead to roll errors and potential failed maneuver.

Closes #242.

In that issue a number of 1.0 arcsec was discussed, but I think it is OK to expand to 1.25 arcsec and reduce the number of stars excluded from about 800 to about 90.

This passes unit tests.  Was not planning any separate validation.

### DIFF
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -354,7 +354,7 @@ class GuideTable(ACACatalogTable):
               (stars['mag_err'] < 1.0) &  # Mag err < 1.0 mag
               (stars['ASPQ1'] < 20) &  # Less than 1 arcsec offset from nearby spoiler
               (stars['ASPQ2'] == 0) &  # Proper motion less than 0.5 arcsec/yr
-              (stars['POS_ERR'] < 3000) &  # Position error < 3.0 arcsec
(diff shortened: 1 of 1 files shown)


taldcroft:
> Good to bump the warning in sparkles from 1.00 to 1.25 then?

Yes.

taldcroft:
I'll do the sparkles update.

jeanconn:
Thanks!

PR #274, opened 2019-02-15
Clip mica dark current map to exclude negative values

### LINKED ISSUE #265: What should we do with pixels with large negative dark current?
For star selection and review how should a pixel with a reported value like the -8570 below be handled?   Is it a bad/broken pixel that should just be completely avoided for guide stars and fids?   Or should it just be set to 0 so it doesn't cause numeric issues?

```
ipdb> img
<ACAImage row0=-215 col0=339
array([[  142,   147,    12,    42,    14,    13,    21,   190],
       [   18,    34,    36,   524,    83,    12,    28,    26],
       [ -408, -8570,    26,    42,    29,   126,    26,    25],
       [  179,   183,    85,    68,   359,    21,    24,    17],
       [  187,     0,    38,   612,    30,    27,     8,     9],
       [   27,   -12,    17,    86,    35,   153,    19,   133],
       [   16,    57,    11,    16,    18,    27,    62,    27],
       [   27,   205,    27,    18,    69,    15,    23,     1]])>
```

taldcroft:
This should be non-controversial.

Closes #265

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -614,6 +614,11 @@ class ACACatalogTable(BaseCatalogTable):
             self.log(f'Getting dark cal image at date={self.date} t_ccd={self.t_ccd:.1f}')
             self.dark = get_dark_cal_image(date=self.date, select='before',
                                            t_ccd_ref=self.t_ccd, aca_image=True)
+
+            # Dark cal images from archive can have negative values (sometimes large)
+            # but this is not physical.  Do an in-place lower-limit clip to 0.
+            np.clip(self.dark, a_min=0, a_max=None, out=self.dark)
+
         elif not isinstance(self.dark, ACAImage):
             self.dark = ACAImage(self.dark, row0=-512, col0=-512, copy=False)
 


proseco/core.py
```diff
@@ -614,6 +614,11 @@ def set_attrs_from_kwargs(self, **kwargs):
             self.log(f'Getting dark cal image at date={self.date} t_ccd={self.t_ccd:.1f}')
             self.dark = get_dark_cal_image(date=self.date, select='before',
                                            t_ccd_ref=self.t_ccd, aca_image=True)
+
+            # Dark cal images from archive can have negative values (sometimes large)
+            # but this is not physical.  Do an in-place lower-limit clip to 0.
+            np.clip(self.dark, a_min=0, a_max=None, out=self.dark)
+
```
jeanconn: I can't find a way to make this controversial, but I don't recall if there is operational code is passing the already-fetched dark map, and would really need to clip it after this `if` block.

I can confirm that this change makes the warning go away that came up in products review on obsid 48430 with that pixel with large negative value.

taldcroft: I thought about that, and in the end it seems if the user passes in a dark current map, that's what should be used.  I then went to change the bit that sets bad pixels (even for a user-supplied map) but that broke tests so I stopped.  Probably want to come back to that someday.

jeanconn: OK. And yeah, I thought that as well, just couldn't remember if all of the use cases for passing the dark map were from the user.  If negative values in the dark map were a big issue, I'd suggest a warning after this block, but seems like it is probably fine without one.

taldcroft: Well there is one critical case ORviewer and we know that does not pass it.  I can't think of any other code specifically calling dark_cal.get_dark_cal_image() and passing that.  There is no reason we (or at least I) would write that when get_aca_catalog does the same.  (famous last words?)

jeanconn: From a search in proseco, it looks like the only other independent get_dark_cal_image calls are in the reporting.  So do we want to clip it there too or leave it alone (or clip in mica)?

taldcroft:
Plan B is https://github.com/sot/mica/pull/189.  That might just be the thing to do.

taldcroft:
We're doing this in mica!

Issue #259, opened 2019-02-06
Review imposter thresholds


jeanconn:
Remove critical aca_preview warnings by either changing stage 5 imposter threshold or changing warning threshold.

Threshold should be both consistent and supported by some analysis.

PR #280, opened 2019-02-19
Move dark attribute from MetaAttribute to property

taldcroft:
This simplifies code, and makes the "dark" attribute a reliably-available property.

It also results in the AcqTable object using the "correct" dark current map based on `t_ccd_acq`.

### DIFF
--- a/proseco/catalog.py
+++ b/proseco/catalog.py
@@ -114,9 +114,6 @@ def _get_aca_catalog(**kwargs):
     aca.set_stars(filter_near_fov=False)
     kwargs.pop('stars', None)
 
-    # Share a common dark map for memory / processing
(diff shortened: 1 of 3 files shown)


jeanconn:
At this point, are we provisionally done with proseco 4.4?

taldcroft:
Has #259 been addressed?  In either case I'm fine with cutting 4.4 now.  The only question is about remaining issues in sparkles that might need an upstream proseco fix.

jeanconn:
Right, I wasn't sure if we'd want to recut at 4.4 for anything else with a fix or just call them patches.

Regarding #259, I'd say it has not really been addressed but that the changes we have merged seem reasonable.

PR #281, opened 2019-02-20
Always cast 'att' attribute into Quat object

taldcroft:
Leaving the `att` (attitude) attribute as whatever the user supplied instead of casting to Quat makes life difficult for downstream code.  This fixes that and with this in place one can be sure that `aca.att` is a `Quat`.

This passes tests.  Sparkles passes tests (except for the known fail).

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -332,6 +332,11 @@ class MetaAttribute:
         return (f'<{self.__class__.__name__} name={self.name} default={self.default} '
                 f'pickle={self.pickle}>')
 
+class QuatMetaAttribute(MetaAttribute):
+    def __set__(self, instance, value):
+        if not isinstance(value, Quat):
+            value = Quat(value)
+        instance.meta[self.name] = value
 
 class IntMetaAttribute(MetaAttribute):
     def __set__(self, instance, value):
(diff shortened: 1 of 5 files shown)


proseco/core.py
```diff
@@ -332,6 +332,11 @@ def __repr__(self):
         return (f'<{self.__class__.__name__} name={self.name} default={self.default} '
                 f'pickle={self.pickle}>')
 
+class QuatMetaAttribute(MetaAttribute):
+    def __set__(self, instance, value):
+        if not isinstance(value, Quat):
+            value = Quat(value)
```
jeanconn: This seems fine.  I've sometimes run into issues with the not-quite-normalized Quat, but hopefully anything that reaches this point will be fine.

taldcroft: Default handling of a not-quite-normalized Quat should be handled upstream by Quat.  Maybe auto-fix if within some threshold?  Is there something like that already?

jeanconn: Right now there's just a good error message.   For proseco I had the MATLAB side explicitly normalize to avoid the possibility of it being a problem with these attitudes (relying on later checking to catch any problem).

PR #282, opened 2019-02-20
Fix guide and acq report templates after #281

taldcroft:
Oops.  But the report templates were just fragile before and depended on user supplying attitude as a list of [ra, dec, roll].  This also moves the formatting into the template where it belongs.


jeanconn:
Ah.  Sorry I missed that when checking out #281 ; good catch.

PR #285, opened 2019-02-21
Update acq report format a bit, tidy code, change raise_exc default

taldcroft:
Changing the default of `raise_exc` is the big API change here.

### DIFF
--- a/proseco/tests/test_catalog.py
+++ b/proseco/tests/test_catalog.py
@@ -115,7 +115,7 @@ def test_exception_handling():
                           t_ccd_acq=-10, t_ccd_guide=-10,
                           detector='ACIS-S', sim_offset=0, focus_offset=0,
                           n_guide=8, n_fid=3, n_acq=8,
-                          include_ids_acq=[1])  # Fail
+                          include_ids_acq=[1], raise_exc=False)  # Fail
     assert 'include_ids and include_halfws must have same length' in aca.exception
 
     for obj in (aca, aca.acqs, aca.guides, aca.fids):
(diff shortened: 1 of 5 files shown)


taldcroft:
Hmm, I don't see a way to change the branch this merges into.  Gotta run, but I guess will make a new PR for merge to master?

jeanconn:
For me, I can change target branches from the edit (that looks like it is for the subject, but gives a drop down for target branch).  Is that what you meant?

taldcroft:
But of course!

proseco/tests/test_catalog.py
```diff
@@ -357,7 +357,8 @@ def test_t_ccd_effective_acq_guide(t_ccd_case):
 
 
 def test_call_args_attr():
-    aca = get_aca_catalog(**mod_std_info(optimize=False, n_guide=0, n_acq=0, n_fid=0))
+    aca = get_aca_catalog(**mod_std_info(optimize=False, n_guide=0, n_acq=0, n_fid=0),
+                          raise_exc=False)
```
jeanconn: Not sure about this one.

taldcroft: This is reproducing the original default, and this test requires `raise_exc` to pass.

taldcroft: I meant it needs `raise_exc=False` to pass.

jeanconn: I suppose it really just needs raise_exc to *do* the test on call_args.

PR #290, opened 2019-02-22
Set stars for plot() if needed

jeanconn:
Set stars for plot() if needed

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -691,6 +691,8 @@ class ACACatalogTable(BaseCatalogTable):
         :param kwargs: other keyword args for plot_stars
         """
         kwargs.setdefault('catalog', self.get_catalog_for_plot())
+        if self.stars is None:
+            self.set_stars()
         kwargs.setdefault('stars', self.stars)
         kwargs.setdefault('bad_stars', self.bad_stars_mask)
         return super().plot(ax, **kwargs)


proseco/core.py
```diff
@@ -691,6 +691,8 @@ def plot(self, ax=None, **kwargs):
         :param kwargs: other keyword args for plot_stars
         """
         kwargs.setdefault('catalog', self.get_catalog_for_plot())
+        if self.stars is None:
```
jeanconn: It looks like set_stars has its own logic to try to decide where to *get* the stars, but I think it is relatively low-impact to just have a go at getting something if we reach this point and have no stars.

Unless this should be done from BaseCatalogTable or from chandra_aca.plot itself...

taldcroft:
I wanted to completely re-work the stars attribute handling in a fashion similar to dark, but it turns out to be slightly trickier and I punted.  Did you want this for 4.4?

taldcroft:
But to be clear getting the `stars` attribute automatically working and getting rid of `set_stars()` is a priority.

jeanconn:
Not critical for 4.4, I just wasn't sure if there was any reason *not* to throw something like this in for the time being.  But given that the use case is mostly so I can make debugging plots, there's no rush now that I'm comfortable with just doing
```
aca.set_stars()
aca.plot()
```
on a pickle-loaded catalog as needed.

PR #291, opened 2019-02-23
Use np.ndarray instead of ACAImage for dark map

taldcroft:
The `ACAImage` class is nice and convenient, but sadly even a few lines of pure Python ends up being really slow for array item access.  The change in this PR makes the run time about 75-80% of what it is with current master.  So for 30 obsids with a current average time of 1.1 sec, that would shave off 6-7 seconds out of 33.0.

PR #292, opened 2019-02-24
Improve performance

taldcroft:
This includes #291 and substantially improves performance.  Mostly (sadly) this is finding ways around the poor performance of astropy Table, but the good news is that this introduces no algorithmic changes.

For the 1468 catalogs in the standard acq regression test set, there are no diffs.  In this set the mean time to generate a catalog was reduced from 1.12 sec to 0.51 sec.  At this point the time for ACAReviewTable review is probably significant in the overall performance budget.

### DIFF
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -622,17 +622,20 @@ def check_mag_spoilers(cand_stars, ok, stars, n_sigma):
 
     mag_spoiled = np.zeros(len(ok)).astype(bool)
     rej = []
-    for cand in cand_stars[ok]:
-        pix_dist = np.sqrt(((cand['row'] - stars['row']) ** 2) +
-                           ((cand['col'] - stars['col']) ** 2))
-        spoilers = ((np.abs(cand['row'] - stars['row']) < 10) &
-                    (np.abs(cand['col'] - stars['col']) < 10))
+    cand_idxs = np.flatnonzero(ok)
+
+    for cand_idx in cand_idxs:
+        cand = cand_stars[cand_idx]
+        spoil_idxs = np.flatnonzero(
+            (np.abs(cand['row'] - stars['row']) < 10) &
+            (np.abs(cand['col'] - stars['col']) < 10))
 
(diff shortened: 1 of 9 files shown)


proseco/guide.py
```diff
@@ -623,17 +622,20 @@ def check_mag_spoilers(cand_stars, ok, stars, n_sigma):
 
     mag_spoiled = np.zeros(len(ok)).astype(bool)
     rej = []
-    for cand in cand_stars[ok]:
-        pix_dist = np.sqrt(((cand['row'] - stars['row']) ** 2) +
```
taldcroft: Don't need to calculate the distance to every star, only stars that are in the spoiler box.  (Right?)

jeanconn: Sure.  That seems right.  I'd tried to optimize by just not calling the full routine much, but plenty of room to improve it locally.  Thanks!

proseco/guide.py
```diff
@@ -622,17 +622,20 @@ def check_mag_spoilers(cand_stars, ok, stars, n_sigma):
...
-                           ((cand['col'] - stars['col']) ** 2))
-        spoilers = ((np.abs(cand['row'] - stars['row']) < 10) &
-                    (np.abs(cand['col'] - stars['col']) < 10))
+    cand_idxs = np.flatnonzero(ok)
+
+    for cand_idx in cand_idxs:
+        cand = cand_stars[cand_idx]
+        spoil_idxs = np.flatnonzero(
+            (np.abs(cand['row'] - stars['row']) < 10) &
```
jeanconn: Though I suppose this does calculate the distance to every star (right?) but at least doesn't also square and square root it.

jeanconn:
I can't tell from the PR text @taldcroft had benched this on linux, but I can report I am seeing the advertised 2x improvement in aggregate on that platform (as in it just took me half as long to run my separate regression test set on kadi).

taldcroft:
I never run on linux if I can avoid it.  :smile:

PR #294, opened 2019-02-25
Fix bug in setting stars attribute in reporting

taldcroft:
In this code the `stars` attribute was always being set even if already set, but without calling the `near_fov_filter` filtering. So there got to be a mismatch between `stars` and `bad_stars_mask`.

A small mystery to me is why this wasn't failing more.  It might have been that it worked as long as none of the stars in the small acq box plot were bad.


jeanconn:
Thanks @taldcroft !  Regarding "A small mystery to me is why this wasn't failing more. It might have been that it worked as long as none of the stars in the small acq box plot were bad."  I was having the same thought.

taldcroft:
Thanks for reviews @jeanconn !

PR #296, opened 2019-02-26
 Remove use of test(get_version=True)

### LINKED ISSUE #295: remove test() calls in release code
https://github.com/sot/proseco/blob/8721ee3c94756f5abb8e4dc7e0a0c381cec7e1a0/proseco/catalog.py#L15

This code causes a dialog to flash in Windows, which is undesirable for the release.

taldcroft:
Merging to 4.4.x branch when approved, so this should be cherry-picked into master as well.

Passes unit tests.  Functional test (since there is no explicit unit test for version):
```
In [1]: from proseco import get_aca_catalog
In [2]: aca = get_aca_catalog(20603)
In [3]: aca.version
Out[3]: '4.4.1'
```

Fixes #295

### DIFF
--- a/proseco/__init__.py
+++ b/proseco/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "4.4"
+__version__ = "4.4.1"
 
 
 def get_aca_catalog(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


taldcroft:
Applied to master with ` git cherry-pick -m 1 1502441`.

PR #297, opened 2019-03-21
Allow optimizing acq halfw for included star

taldcroft:
This allows proseco to optimize the box size (halfw) for a star that is force included.  This is done by setting include_halfw=0.
```
In [13]: aca = get_aca_catalog(20603, include_ids_acq=[116923672, 40112304, 40113544], include_halfws_acq=[0, 0, 0])

In [14]: aca
Out[14]: 
<ACATable length=11>
    0     1         2  FID  8x8   0.000    7.00    8.00  -773.20 -1742.03     1     1    25
    1     2         4  FID  8x8   0.000    7.00    8.00  2140.23   166.63     1     1    25
    2     3         5  FID  8x8   0.000    7.00    8.00 -1826.28   160.17     1     1    25
    3     4  40112304  BOT  6x6   0.689    9.79   11.29 -1644.35  2032.47    20     1   160
    4     5  40113544  BOT  6x6   0.981    7.91    9.41   102.74  1133.37    20     1   160
    5     6  40114416  BOT  6x6   0.919    9.75   11.25   394.22  1204.43    20     1   140
    6     7 116791824  BOT  6x6   0.968    9.15   10.65   622.00  -953.60    20     1   160
    7     8 116923496  ACQ  6x6   0.963    9.14   10.64 -1337.79  1049.27    20     1   120
    0     9 116923528  ACQ  6x6   0.586    9.84   11.34 -2418.65  1088.40    20     1   160
    1    10 116791744  ACQ  6x6   0.606   10.31   11.81   985.38 -1210.19    20     1    60
    2    11 116923672  ACQ  6x6   0.126   10.97   12.47 -2307.80  1442.43    20     1    60
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -62,6 +62,12 @@ exclude_ids_guide   int or list of AGASC IDs of stars to exclude from guide cata
 stars               table of AGASC stars (will be fetched from agasc if None)
 =================== ===============================================================
 
+Within the ``include_halfws_acq`` list, one can supply the value ``0`` for a
+star instead of a typical legal value such as ``60`` or ``120``.  In that case
+proseco will run the normal optimization and choose the best halfwidth for that
+included star.  If the ``include_halfws_acq`` argument is not supplied or set
+to ``[]`` then all halfwidths will be chosen by proseco.
+
 **Debug**
 
 ============== =========================================================
--- a/proseco/acq.py
+++ b/proseco/acq.py
@@ -136,6 +136,11 @@ class AcqTable(ACACatalogTable):
     include_halfws = AliasAttribute()
     exclude_ids = AliasAttribute()
 
+    # IDs that are included but with halfw=0 which implies to optimize halfw
+    # instead of freezing at the provided value.  This attribute is set internally
+    # based on the values of include_halfws.
+    include_opt_ids = MetaAttribute(is_kwarg=False)
+
     p_man_errs = MetaAttribute(is_kwarg=False)
     cand_acqs = MetaAttribute(is_kwarg=False)
     p_safe = MetaAttribute(is_kwarg=False)
@@ -459,6 +464,10 @@ class AcqTable(ACACatalogTable):
         # Ensure values are valid box_sizes
         grid_func = interp1d(ACQ.box_sizes, ACQ.box_sizes,
                              kind='nearest', fill_value='extrapolate')
+        self.include_opt_ids = [acq_id for acq_id, halfw,
+                                in zip(self.include_ids, self.include_halfws)
+                                if halfw == 0]
+
         self.include_halfws = grid_func(self.include_halfws).tolist()
 
         super().process_include_ids(cand_acqs, stars)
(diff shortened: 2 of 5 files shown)


taldcroft:
@jskrist - let me know if this works for you and would let you chop more code in SAUSAGE.  Basically for a hand-edit in that case you supply 0 as the box size halfwidth.  And of course I'm not tied to that API, just picked something quickly that I knew would work with MATLAB.  In a Python-only world I would use `None` instead of `0`.

jskrist:
@taldcroft - That looks good to me.  I can send a `None` to Python, however, I would want to be guaranteed that I would never receive a `None` from Python after the algorithm has run, since the pyexec MATLAB comand cannot convert `NoneType` variables.

taldcroft:
@jeanconn @jskrist - we should consider getting this into Ska3 for the next MATLAB release.  Allowing proseco to choose the best acq star halfwidth should really be the default in operations when a star is force-included.  This change makes it fairly easy, either by supplying all zero values or else sending over an empty list.  Is this possible right now in the MATLAB interface?

jskrist:
@taldcroft - the MATLAB interface allows for the sending of zero values and empty lists.  I will have to look into the code a bit more to determine how much needs to be changed on the MATLAB side to start using this update, but it seems pretty straight forward.

taldcroft:
So if somebody just typed one or more zeros into the ORviewer field for halfwidths, would that do the trick?  (Without any software change beyond this one).

jskrist:
It looks like the halfwidths can be set in ORViewer when forcing a specific star to be included, and the keyword being passed into proseco for this is currently `include_halfws_acq`, so a user could specify that as `0` and it would be passed into proseco without any code changes. 👍

jskrist:
@taldcroft  - I pulled these changes into my local python dev branch for the MATLAB FOT Tools, and ran into the issue, that I don't have the 'ska_helpers' package.  I will get that package as well, and keep testing, but it is worth noting that this is a dependency that will need to come along with these changes into the MATLAB FOT Tools.

jeanconn:
Thanks James.  I'm starting to track the ska3-matlab requirements in https://github.com/sot/skare3/pull/297 .  ska_helpers is in there but I'll draw more attention to it as a new requirement.

jeanconn:
@jskrist I merged this with optimism that it looks fine and if you run into issues with functional testing or weirdness in integration testing we'll just patch and continue.

jskrist:
@jeanconn, @taldcroft  - I have been seeing a strange thing in my testing.  When I take an ACA Catalog found by proseco, and explicitly force some of the stars found to be included as the same thing proseco chose for then in the first place (ACQ or BOT), and then set their half-widths to 0, proseco suddenly decides that their half-widths all need to be 60, and I then start getting critical messages like `critical: P2: 0.94 less than 2.0 for OR`.

If only a "small number" of stars are forced, proseco seems to be finding a reasonable half width, but for more stars, it seems to be defaulting to 60.

taldcroft:
@jskrist - copy.  Do you have an obsid or test catalog to demonstrate?  I'll try to reproduce based on your description in the mean time.

taldcroft:
@jskrist - never mind, I can reproduce this easily.

PR #301, opened 2019-06-10
Implement capping of catalog MAXMAG to 11.2 mag

taldcroft:
Based on discussions in emails ("maxmag" near June 7, 2019), the MAXMAG parameter in commanded star catalogs should be capped at 11.2 mag.  This makes that change in proseco catalogs.

This PR does **not** implement capping of the `maxmag` parameter found in the acq star selection code.  That value of `maxmag` is actually a different quantity, defined as 1-sigma above the star mag.  In practice that will rarely be above 11.2 for selected stars, but in any case I would rather leave the acq selection algorithm unchanged so this PR does not introduce any change in star selection.

cc: @mbaski @jskrist

### DIFF
--- a/proseco/acq.py
+++ b/proseco/acq.py
@@ -614,7 +614,7 @@ class AcqTable(ACACatalogTable):
         # Imposters
         ext_box_size = box_size + dither
         kwargs = dict(star_row=acq['row'], star_col=acq['col'],
-                      maxmag=acq['mag'] + acq['mag_err'],  # + 1.5, TO DO: put to characteristics
+                      maxmag=acq['mag'] + acq['mag_err'],
                       box_size=ext_box_size,
                       dark=dark,
                       bgd=bgd,  # TO DO deal with this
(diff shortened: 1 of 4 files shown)


jeanconn:
Should this just be cut as proseco 4.5?  I don't think there are any doc or other action items to close out with regard to the new maxmag.

taldcroft:
I'm going to defer to your judgement as Ska release manager.  I'm honestly not sure about the MATLAB release schedule, if this should go in to the Matlab release, or when, and etc.  The handling of this is not entirely clear cut as you probably realized.

jeanconn:
OK.  I think we should just cut this as a release, but I didn't consider the maxmag change as a bugfix.  This would mean a 4.5 release; I think we haven't really decided if that means we need to do anything with 4.3.x and 4.4.x branches.

jeanconn:
Oh and I suppose we need to update the ACA load review checklist ACA-041 and also update the starcheck check to match .

PR #303, opened 2019-06-30
Store the dark current date in ACACatalogTable object

### LINKED ISSUE #200: Include provenance information in pickle

taldcroft:
Partially addresses #200.

### DIFF
--- a/proseco/catalog.py
+++ b/proseco/catalog.py
@@ -116,6 +116,10 @@ def _get_aca_catalog(**kwargs):
     aca.log('Starting get_acq_catalog')
     aca.acqs = get_acq_catalog(stars=aca.stars, **kwargs)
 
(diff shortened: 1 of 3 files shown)


jeanconn:
If the date is stored, should it be used when making the reports?  I don't remember which dark cal is used from the reports if made from a pickle (and can there be inconsistencies now?).

taldcroft:
Maybe, not sure.  This PR is not trying to make the entire system perfect, just a little better.

jeanconn:
Sure.  I just couldn't figure it out when I took a quick look and figured it was something that should be clear.  I can open a separate issue.

PR #307, opened 2019-08-21
Use -8.8C for planning limit and effective t_ccd determination

jeanconn:
Use -8.8C for planning limit and effective t_ccd determination

### DIFF
--- a/proseco/characteristics.py
+++ b/proseco/characteristics.py
@@ -29,8 +29,9 @@ max_ccd_col = CCD['col_max'] - CCD['col_pad']  # Max allow col for stars (SOURCE
 col_spoiler_mag_diff = 4.5
 col_spoiler_pix_sep = 10  # pixels
 
-# ACA T_ccd planning limit (degC)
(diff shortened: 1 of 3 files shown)


proseco/characteristics.py
```diff
@@ -30,7 +30,7 @@
 col_spoiler_pix_sep = 10  # pixels
 
 # ACA T_ccd planning limit (degC)
-aca_t_ccd_planning_limit = -9.5
+aca_t_ccd_planning_limit = -8.8
```
taldcroft: I was planning a two-fer here to also change that name to `aca_t_ccd_penalty_limit = -8.8`.  Not sure if other tools use the `aca_t_ccd_planning_limit` name.  We could (if necessary) add it back in at the current value of `-7.8`.

The point of doing the renaming is that things are confusing enough as it stands, so having mis-named values in the code is no good.

jeanconn: OK.  It looks like it is called directly in a sparkles test so would need to fix there too.

PR #309, opened 2019-09-01
WIP: Add a force_catalog keyword arg for analysis

jeanconn:
Add a force_catalog keyword arg for analysis

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -549,6 +549,7 @@ class ACACatalogTable(BaseCatalogTable):
     include_ids_guide = IntListMetaAttribute(default=[])
     exclude_ids_guide = IntListMetaAttribute(default=[])
     optimize = MetaAttribute(default=True)
+    force_catalog = MetaAttribute(default=False)
     verbose = MetaAttribute(default=False)
     print_log = MetaAttribute(default=False)
     log_info = MetaAttribute(default={}, is_kwarg=False)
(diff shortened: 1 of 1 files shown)


jeanconn:
This has been hanging around and needs at least a test.

taldcroft:
I'd rather not put this in the API for this rarely-used option.  Note that in addition to a test you would need to update all the docs that reference available keywords.

I'd suggest instead to add a function to `core.py` : `includes_from_obsid(obsid)` that returns a dict so that you can do:
```
from proseco.core import includes_from_obsid
get_aca_catalog(obsid, **includes_from_obsid(obsid))
```

jeanconn:
I was thinking that get_aca_catalog(obsid, force_catalog=True) might be more useful than get_aca_catalog(obsid) at this point.

taldcroft:
What are the use cases for that?  I think all the existing unit tests that reference `obsid` are correct as written, and in the recent case for SEP0919 prelim I did not see a need to look at the force-catalog version of the 3 reports you supplied.

jeanconn:
I wasn't suggesting changing the default (which would apply to the "I think all the existing unit tests that reference obsid are correct as written").  I was much more interested in updating the acq stats database (or a new table) to have the last 5 years of proseco P2 values and acq probs for individual stars to do something like https://github.com/sot/starcheck/pull/257#discussion_r224481777 .

Which is, of course, fine with the helper function idea you suggest.  My point was just that I am *also* not using get_aca_catalog(obsid) much at this point, though of course we may want to do so for large recalibration efforts.

taldcroft:
Good points, but not a driver for an API change in my view.

jeanconn:
Closing this in favor of #311

PR #310, opened 2019-09-14
PEP8, flake8 and sphinx RST fixes

taldcroft:
This is the beginning of the end of PEP8 and flake8 violations in Ska3 code.  I'm using the following for my `~/.config/pycodestyle` file (which also controls `autopep8`):
```
[pycodestyle]
max-line-length = 100
```
I've applied for [GitHub actions](https://github.com/features/actions) for the `sot` org.  When we get accepted then we'll start enforcing `flake8` compliance for merge.  E.g. https://github.com/actions/starter-workflows/blob/master/ci/python-package.yml.


proseco/acq.py
```diff
@@ -9,7 +9,6 @@
 import numpy as np
 from scipy import ndimage, stats
 from scipy.interpolate import interp1d
-from astropy.table import Table
```
javierggt: I am assuming that this was imported and not used.

taldcroft: Correct.

taldcroft: I did run tests, but it isn't a bad idea as reviewer to just independently check out the repo in this branch and run `python setup.py test` or `pytest proseco`.

PR #317, opened 2020-02-03
Fixes for shiny Ska3

taldcroft:
## Description

This has some changes that are required to pass tests in ska3-shiny (Python 3.6 version).  Details are given in the commit messages.

This branch passes tests in current ska3-flight.  The only catch is that 186897d will be slower in ska3-flight.

- [x] Passes unit tests on MacOS (ska3-flight)
- [x] Passes unit tests on MacOS (shiny: astropy 4.0.1, numpy 1.18, etc)
- [x] Timing regression testing acceptable

### Timing

Running `proseco/tests/timing.py : time_get_aca_catalog(n_samples=100)`

- ska3-flight : master (4.8.1 @ 52b75ee):  370 ms / catalog
- ska3-flight : shiny-fix (6dd88ea): 388 msec / catalog (5% slower)
- shiny : shiny-fix (6dd88ea): 317 msec / catalog (14% faster)

So this is a little slower in current Ska3-flight, but will be faster in shiny.

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -507,7 +507,8 @@ class BaseCatalogTable(Table):
         kwargs.setdefault('bad_stars', np.zeros(len(kwargs['stars']), dtype=bool))
 
         fig = plot_stars(attitude=self.att, ax=ax, **kwargs)
-        plt.show()
+        if 'agg' not in plt.get_backend().lower():
+            plt.show()
 
         return fig
 
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 *.pkl
 *.npz
 /validate
+/obs?????
+/.idea
+/.vscode
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/proseco/acq.py
+++ b/proseco/acq.py
@@ -628,7 +628,8 @@ class AcqTable(ACACatalogTable):
         acqs_init = cand_acqs[acq_indices]
 
         # Transfer to acqs (which at this point is an empty table)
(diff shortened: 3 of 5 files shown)


jeanconn:
@taldcroft do you want this for the matlab release or don't care?

taldcroft:
Not for the MATLAB release.  Still need to check on performance and this is a lower priority.

jeanconn:
That's what I figured.  So I think proseco, chandra_aca, sparkles are good to have releases cut now.

taldcroft:
@jeanconn - I believe this is ready now.  I've done performance testing and things look OK, see the description.

proseco/core.py
```diff
@@ -507,7 +507,8 @@ def plot(self, ax=None, **kwargs):
         kwargs.setdefault('bad_stars', np.zeros(len(kwargs['stars']), dtype=bool))
 
         fig = plot_stars(attitude=self.att, ax=ax, **kwargs)
-        plt.show()
+        if 'agg' not in plt.get_backend().lower():
+            plt.show()
```
taldcroft, line 511: In shiny (new matplotlib) one gets a warning in tests about calling `plt.show()` for a non-interactive backend.

taldcroft:
I think that testing individual package updates by hand will not save trouble (i.e. reduce level of effort). Windows-only fails are pretty rare, so until we get to a point where tests are automated or really easy to run, it will be less effort to fix rare problems at the end than always test up front.

Note that despite efforts in that direction, Windows testing is a bit of a pain and we are not testing the exact environment that the FOT will be using anyway.

PR #320, opened 2020-02-07
Update tests for 2020-02 acq model update

jeanconn:
Update tests for 2020-02 acq model update

This updates the regression data for the new acq model and cuts the tests of the deprecated thumbs_up property.


jeanconn:
I could have done something nicer (like improved the tests) but given how often we update the acq model this still seems fine.

PR #324, opened 2020-04-23
Keep box_sizes in the same order as the acq_indices

jeanconn:
## Description

Keep box_sizes in the same order as the acq_indices

## Testing

- [x] Passes unit tests on linux (of MacOS, linux, Windows (at least one required))
- [x] Functional testing (added new test of the bug as seen in the wild, apparently [Add a test for Jean](https://github.com/sot/proseco/commit/2d15cda96b50961d14bb06f41ba5d0953d20cadb) didn't get us all the way there)

Fixes #323

PR #325, opened 2020-04-24
Fix bug in halfws for force-include acq stars

### LINKED ISSUE #323: Included halfws may not line up with include ids
In APR2720A in obsid 47250, the starcheck process passed these args to proseco for catalog review

```
{'att': array([ 0.37484371, -0.40057847, -0.79053942,  0.27216999]),
 'date': '2020:121:04:31:50.888',
 'detector': 'ACIS-S',
 'dither_acq': <ACABox y=20.00149630449384 z=20.00149630449384>,
 'dither_guide': <ACABox y=20.00149630449384 z=20.00149630449384>,
 'focus_offset': 0,
 'include_halfws_acq': [140, 160, 160, 160, 160, 160, 160, 160],
 'include_ids_acq': [810423632,
  810428640,
  812125920,
  812130960,
  812131816,
  886441184,
  810427792,
  810428880],
 'man_angle': 91.51488826,
 'n_acq': 8,
 'n_fid': 0,
 'n_guide': 0,
 'obsid': 0,
 'sim_offset': 0,
 't_ccd_acq': -10.636981637514856,
 't_ccd_guide': -10.636981637514856}
```

Halfw 140 is supposed to be associated with 810423632, but ends up on 810427792

```
In [15]: aca
Out[15]: 
<ACATable length=8>
 slot  idx      id    type  sz   p_acq  ...   yang     zang    dim   res  halfw
int64 int64   int32   str3 str3 float64 ... float64  float64  int64 int64 int64
----- ----- --------- ---- ---- ------- ... -------- -------- ----- ----- -----
    0     1 810423632  ACQ  8x8   0.000 ...  1325.38  2403.91    20     1   160
    1     2 810427792  ACQ  8x8   0.979 ...  -715.07  2085.91    20     1   140
    2     3 812131816  ACQ  8x8   0.981 ...  1363.16 -1859.97    20     1   160
    3     4 886441184  ACQ  8x8   0.981 ... -2093.64   162.33    20     1   160
    4     5 812125920  ACQ  8x8   0.970 ...   362.73 -2371.63    20     1   160
    5     6 810428880  ACQ  8x8   0.955 ...   888.47   922.52    20     1   160
    6     7 812130960  ACQ  8x8   0.953 ...  1219.38   161.94    20     1   160
    7     8 810428640  ACQ  8x8   0.951 ...    91.02   865.90    20     1   160
```

This is more a problem for starcheck than operations.

taldcroft:
## Description

This fixes #323 and supercedes #324. The actual bug fix is simple but I spent some time understanding how it can come about and why the original tests did not show a problem.  In short the problem comes up when the initial `p_acq` for included candidate stars does not increase monotonically with decreasing `mag`.  This can happen because of a spoiler or hot pixel.

This only impacts cases with more than one included star.  In actual star selection with MATLAB choosing more than one star is not common and typical practice now would be to not explicitly select the halfwidths.  It has a definite impact in starcheck which uses force-include of all stars and halfwidths for every catalog.

## Testing

- [x] Functional testing

Also ran this test from #324 and confirmed that the update changes this test from fail to pass, but did not include in the final PR.
```
def test_obsid_47250():
    """Test include all 8 acq ids has right halfws for right stars"""

    # This uses the attitude/stars from obsid 47250 in APR2720
    att = [0.37484371, -0.40057847, -0.79053942, 0.27216999]
    stars = StarsTable.from_agasc(att, date='2018:230')
    dark = DARK40.copy()
    kwargs = mod_std_info(att=att, stars=stars, dark=dark,
                          n_guide=8, n_fid=0, n_acq=8, man_angle=90)
    kwargs['include_halfws_acq'] = [140, 160, 160, 160, 160, 160, 160, 160]
    kwargs['include_ids_acq'] = [810423632,
                                 810428640,
                                 812125920,
                                 812130960,
                                 812131816,
                                 886441184,
                                 810427792,
                                 810428880]
    aca = get_aca_catalog(**kwargs)

    # Confirm that the included stars have the assigned halfws
    for id, halfw in zip(kwargs['include_ids_acq'],
                         kwargs['include_halfws_acq']):
        assert aca.acqs.get_id(id)['halfw'] == halfw
```

Fixes #323
Closes #324

### DIFF
--- a/proseco/acq.py
+++ b/proseco/acq.py
@@ -603,7 +603,7 @@ class AcqTable(ACACatalogTable):
                     # Find the position in box_sizes that corresponds to include_id
                     # and set to the specified include_halfw.
                     for idx in range(len(acq_indices)):
-                        if include_id == cand_acqs[idx]['id']:
+                        if include_id == cand_acqs[acq_indices[idx]]['id']:
                             box_sizes[idx] = include_halfw
                             break
 
--- a/proseco/tests/test_acq.py
+++ b/proseco/tests/test_acq.py
@@ -1152,3 +1152,30 @@ def test_acq_include_ids_all_halfws_full_catalog():
     assert np.all(aca2.acqs['halfw'] == aca.acqs['halfw'])
 
     assert aca.acqs.calc_p_safe() == aca2.acqs.calc_p_safe()
+
+
+def test_acq_include_ids_all_halfws_full_catalog_with_spoiler():
+    """Test for all 8 acq_ids provided with all include_halfws provided.
+
+    Importantly include stars for which the p_acq is not increasing in the
+    same order as mag. This is the root cause of #323.  For this test two
+    bright acq stars with a strong spoiler are included.
+    """
+    stars = StarsTable.empty()
+    dark = DARK40.copy()
+    ids = np.arange(100, 108)
+    halfws = [60, 80, 100, 120, 140, 160, 60, 80]
+    mags = [10.6, 10.2, 9.8, 7.0, 9.5, 8.5, 7.4, 7.5]
+
+    stars.add_fake_constellation(mag=mags[:6], n_stars=6, size=2000, id=ids[:6])
+    stars.add_fake_star(yang=0, zang=0, mag=mags[6], id=ids[6])
+    stars.add_fake_star(yang=40, zang=0, mag=mags[7], id=ids[7])
+    stars.add_fake_star(yang=20, zang=0, mag=5.0)  # Strong spoiler for 106 and 107
+
+    kwargs = mod_std_info(stars=stars, dark=dark, n_guide=8, n_fid=0, n_acq=8,
+                          man_angle=30, include_ids_acq=ids, include_halfws_acq=halfws)
+    aca = get_aca_catalog(**kwargs)
+    aca.acqs.sort('id')
+    assert np.all(aca.acqs['id'] == ids)
+    assert np.allclose(aca.acqs['mag'], mags)
+    assert np.all(aca.acqs['halfw'] == halfws)


jeanconn:
The other impact being that, since starcheck force includes all the stars and halfwidths for review, that catalogs in that final review could have incorrect acquisition probabilities due to incorrect internal assignment of box sizes for that assessment.

We'll want to get this promoted on the HEAD side relatively quickly for that use case, but I don't see a big operational impact with proseco 4.8.0 on the FOT side.  However, I also don't see work-around from ORViewer if we actually did need to specify halfwidths for stars that have candidates out of mag order.

taldcroft:
Agreed.  Note that your first impact is technically covered by what I said (multiple stars being force-included), but I edited to call out your point that this does regularly happen.

As mentioned I believe that in current practice the case of specifying multiple stars with explicit halfwidths is no longer going to happen in FOT planning. But in any case this bugfix will be available and it now looks like a smaller MATLAB release is going to happen before the 2019b update.

proseco/acq.py
```diff
@@ -603,7 +603,7 @@ def get_initial_catalog(self):
                     # Find the position in box_sizes that corresponds to include_id
                     # and set to the specified include_halfw.
                     for idx in range(len(acq_indices)):
-                        if include_id == cand_acqs[idx]['id']:
+                        if include_id == cand_acqs[acq_indices[idx]]['id']:
```
jeanconn, line 606: This looks fine.  I got a bit lost in why this section had a double loop to try to find matches but I'm good with the more minimal change.

proseco/tests/test_acq.py
```diff
@@ -1152,3 +1152,30 @@ def test_acq_include_ids_all_halfws_full_catalog():
...
+    stars = StarsTable.empty()
+    dark = DARK40.copy()
+    ids = np.arange(100, 108)
+    halfws = [60, 80, 100, 120, 140, 160, 60, 80]
+    mags = [10.6, 10.2, 9.8, 7.0, 9.5, 8.5, 7.4, 7.5]
+
+    stars.add_fake_constellation(mag=mags[:6], n_stars=6, size=2000, id=ids[:6])
+    stars.add_fake_star(yang=0, zang=0, mag=mags[6], id=ids[6])
+    stars.add_fake_star(yang=40, zang=0, mag=mags[7], id=ids[7])
+    stars.add_fake_star(yang=20, zang=0, mag=5.0)  # Strong spoiler for 106 and 107
```
jeanconn, line 1173: And thanks for digging deeper and figuring out how to construct a case with candidates out of mag order!

proseco/acq.py
taldcroft: The double loop was to try keeping the logic simple.  This morning I tried and failed to understand your implementation, brain wasn't firing enough.  But when I looked at the original code the bug and solution was immediately obvious to me.  Not surprising since I wrote the original...

jeanconn: Indeed I think the reason I couldn't quite wrap my head around your implementation was the same "brain firing enough" issue.  I'm still trying to figure out when in the pandemic day I have the most brain power!  Thanks!

Issue #208, opened 2018-12-21
Include / exclude fids and expand MATLAB interface


taldcroft:
In obsid 20268 in AUG2018U, the hand edit reduced the number of fids from 3 to 2 without a corresponding selection of an additional guide star.  This can be worked around, but ends up giving a catalog that wasn't produced by proseco (thus a mismatch between products and the proseco pickle).

In order to continue on the path of having proseco fully choose the catalog, it would be useful to have the option to include/exclude fid lights.  Along those lines, specifying `n_fid`, `n_acq`, and `n_guide` from MATLAB in a GUI would help toward entirely eliminating SAUSAGE hand-edits.

PR #331, opened 2020-07-26
Add include/exclude of fids by index/id

jeanconn:
## Description

Add include/exclude of fids by index/id

I went with the strategy of using the included or excluded ids to reduce the candidate fid sets at the beginning of the fid selection.  If an impossible fid is requested, no candidates will remain and no fids will be selected.  This seemed reasonable for the use cases (though may not work to manually select fid sets used in previous undercover observations with off-CCD fids).

## Testing

- [x] Passes unit tests Linux
- [X] Added tests for functional testing

### DIFF
--- a/proseco/fid.py
+++ b/proseco/fid.py
@@ -178,15 +178,11 @@ class FidTable(ACACatalogTable):
             cand_fid_sets = [fid_set for fid_set in FID.fid_sets[self.detector]
                              if fid_set <= cand_fids_ids]
 
-        if len(self.include_ids_fid) > 0:
-            ok = np.ones(len(cand_fid_sets), dtype=bool)
-            for fidid in self.include_ids_fid:
-                ok = ok & np.array([fidid in fid_set for fid_set in cand_fid_sets])
-            if not np.all(ok):
-                self.log(f'Reducing fid sets from {len(cand_fid_sets)} to '
-                         f'{np.count_nonzero(ok)} via include_ids_fid')
-            cand_fid_sets = [fid_set for idx, fid_set in enumerate(cand_fid_sets)
-                             if ok[idx]]
+        # Restrict candidate fid sets to those that entirely contain the include_ids_set
+        include_ids_set = set(self.include_ids_fid)
+        cand_fid_sets = [fid_set for fid_set in cand_fid_sets if fid_set >= include_ids_set]
+        self.log(f'Reducing fid sets to those that include fid ids {self.include_ids_fid}')
+
         return cand_fid_sets
 
     def set_slot_column(self):
--- a/proseco/fid.py
+++ b/proseco/fid.py
@@ -312,17 +312,25 @@ class FidTable(ACACatalogTable):
 
         self.log(f'Initial candidate fid ids are {cand_fids["id"].tolist()}')
 
-        # Reject candidates that are off CCD, have a bad pixel, or are spoiled
+        # First check that any manually included fid ids are valid by seeing if
+        # the supplied fid is in the initial ids for this detector.
+        if id_diff := set(self.include_ids_fid) - set(cand_fids['id']):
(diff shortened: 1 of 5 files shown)


proseco/fid.py
```diff
@@ -319,7 +334,8 @@ def get_fid_candidates(self):
         for idx, fid in enumerate(cand_fids):
             if (self.off_ccd(fid) or
                     self.near_hot_or_bad_pixel(fid) or
-                    self.has_column_spoiler(fid, self.stars, stars_mask)):
+                    self.has_column_spoiler(fid, self.stars, stars_mask) or
+                    self.is_excluded(fid)):
```
taldcroft, line 338: I think that include_ids should be handled here, which is analogous to acq and guide. If a fid is force-included, then it should be in the selection no matter what, unless it is not valid.  Not-tested code:
```
        if id_diff := set(self.include_ids_fid) - set(cand_fids['id'])
            raise ValueError(f'included fid ids {id_diff} are not valid')

        for fid in cand_fids:
            excluded = (self.off_ccd(fid)
                        or self.near_hot_or_bad_pixel(fid)
                        or self.has_column_spoiler(fid, self.stars, stars_mask)
                        or self.is_excluded(fid))
            included = fid['id'] in self.include_ids_fid

            if not included and excluded:
                idx_bads.append(idx)
```

proseco/tests/test_fid.py
```diff
@@ -212,3 +212,66 @@ def test_fid_hot_pixel_reject():
 
     fids = get_fid_catalog(stars=StarsTable.empty(), dark=dark, **STD_INFO)
     assert fids['id'].tolist() == [2, 3, 6]
+
+
+def test_fids_include_exclude():
+    """
+    Test include and exclude stars for guide.  This uses a catalog with 11 stars:
```
taldcroft, line 219: Looks like copy/paste from another test, needs fixing.

proseco/fid.py
jeanconn: My concern with that is that I don't think sparkles is doing near pixel or column spoiler checks on fids,  so if we allow override in this way (yes, analogous to acq/fid) to be strictly correct we'd also have to implement the checks in sparkles.  My approach of just reducing the candidates was somewhat weak but benign in this way (if no combinations existed, with the included fid and all fids were excluded, at least you'd get the "no fids" warning out of sparkles).

But I suppose we can just add the checks in sparkles (at least I'm pretty sure they aren't in there now).

taldcroft: Adding those checks to sparkles (and starcheck?) would be a good thing, but not strictly required for this PR. 99.9% of the use for forcing fids will be by starcheck, and there you really need to force the exact same catalog as in loads.

jeanconn: I don't have any walrus operator experience, but that first test seems not to work.   Does it need to be explicitly checked == 0 or what is your preference?

jeanconn: Never mind.  Seems to work now.

proseco/fid.py
```diff
@@ -301,16 +312,26 @@ def get_fid_candidates(self):
...
 
-        # Reject candidates that are off CCD, have a bad pixel, or are spoiled
+
+        # First check that any manually included fid ids are valid by seeing if
+        # the supplied fid is in the initial ids for this detector.
+        if id_diff := set(self.include_ids_fid) - set(cand_fids['id']):
+            raise ValueError(f'included fid ids {id_diff} are not valid')
+
+        # Then reject candidates that are off CCD, have a bad pixel, are spoiled,
+        # or are manually excluded, unless the candidates are forcee/manually included.
```
taldcroft, line 322: Typo with `forcee`.

taldcroft: Oh, and I guess a rebase on the flake8 fix.

jeanconn:
This does not fix #208 but should be considered part of that work.

PR #332, opened 2020-07-28
Fix inadvertent re-use of base class `allowed_kwargs`

### LINKED ISSUE #330: Importing fid tests directly gives unexpected keyword argument error
It looks like one can't import the fid tests without importing the acq tests first?  Or some other issue with the attributes.

Importing the fid tests gives error

```
ska3-fido$ python -c "import proseco.tests.test_fid"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/proj/sot/ska/jeanproj/git/proseco/proseco/tests/test_fid.py", line 12, in <module>
    FIDS = get_fid_catalog(stars=StarsTable.empty(), **STD_INFO)
  File "/proj/sot/ska/jeanproj/git/proseco/proseco/fid.py", line 50, in get_fid_catalog
    fids.set_attrs_from_kwargs(obsid=obsid, **kwargs)
  File "/proj/sot/ska/jeanproj/git/proseco/proseco/core.py", line 616, in set_attrs_from_kwargs
    raise ValueError(f'unexpected keyword argument "{name}"')
ValueError: unexpected keyword argument "t_ccd"
```
Importing the acq tests first, no error.

```
ska3-fido$ python -c "import proseco.tests.test_acq; import proseco.tests.test_fid"
ska3-fido$ 
```

taldcroft:
## Description

This fixes #330, primarily by 82b3910 which allows setting something that is a class property even if it is not in the `allowed_kwargs` set.

However, along the way I discovered a more serious problem which was that the class inheritance from `ACACatalogTable` was re-using (for FidTable and ACATable) the base class `allowed_kwargs`. This was the source of the ordering issue in #330 because `FidTable` was trying to set `t_ccd` but relying on `t_ccd` showing up in `allowed_kwargs` because of importing `catalog` (which defined `ACACatalogTable` and added `t_ccd` to `allowed_kwargs`). So I fixed this problem by making sure each class has a private copy of `allowed_kwargs`.

## Testing

- [x] Functional testing : TO DO add a new unit test.

Fixes #330

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -16,4 +16,4 @@ jobs:
     - name: Lint with flake8
       run: |
         pip install flake8
-        flake8 . --count --exclude=docs/conf.py --ignore=W504,E402,F541 --max-line-length=100 --show-source --statistics
+        flake8 . --count --exclude=docs/conf.py --ignore=W503,W504,E402,F541 --max-line-length=100 --show-source --statistics
--- a/proseco/acq.py
+++ b/proseco/acq.py
@@ -117,6 +117,10 @@ class AcqTable(ACACatalogTable):
     """
     Catalog of acquisition stars
     """
+    # Define base set of allowed keyword args to __init__. Subsequent MetaAttribute
+    # or AliasAttribute properties will add to this.
+    allowed_kwargs = ACACatalogTable.allowed_kwargs.copy()
+
     # Catalog type when plotting (None | 'FID' | 'ACQ' | 'GUI')
     catalog_type = 'ACQ'
 
--- a/proseco/catalog.py
+++ b/proseco/catalog.py
@@ -182,6 +182,10 @@ class ACATable(ACACatalogTable):
     as attributes and other methods relevant to the merged catalog.
 
     """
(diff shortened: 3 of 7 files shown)


jeanconn:
Thanks @taldcroft .  This helps.  My brain is still trying to wrap around the bootstrapping it was doing before.  I see you've added a test; does that wrap up the action items?

taldcroft:
The `FidTable` class was getting the `allowed_kwargs` attribute as a reference from the base `ACACatalogTable`. When a `MetaAttribute` or `AliasAttribute` attribute is defined, it puts that name into `allowed_kwargs`. Now `FidTable` has a `t_ccd` property, but in `set_attrs_from_kwargs` it was only checking `allowed_kwargs` to see if a kwarg `t_ccd` could be passed into `__init__`. So when you only import `fid.py` there was no `t_ccd` in `allowed_kwargs`.  BUT, if you imported `acq.py` first, the `AcqTable` has a `t_ccd` `AliasAttribute` and that put `t_ccd` into the base `allowed_kwargs`, at which point the `t_ccd` property setter could be called.

Yup, it is a bit twisted.

PR #333, opened 2020-07-28
Remove defunct thumbs_up property

### LINKED ISSUE #202: Thumbs-up rules for ERs
For ERs there is the freedom to choose the attitude (subject to a host of other constraints), so ideally we'd like to maintain strong ACA catalogs for these.

For instance:

- P2 > 5.0 for ACQ?  This begs the question of how to obtain that apriori and how difficult it will be.  I can look at the standard validation obsids to see how many don't pass.
- 3 stars brighter than 9th" (scaled for temperature with T=-10C as reference) for guide

image

### LINKED ISSUE #275: Cut defunct thumbs_up properties

taldcroft:
## Description

The thumbs-up concept has been replaced by sparkles and it is no longer used.

Need to verify that this is also not used in MATLAB tools ORviewer. An old version that we have in our repo appears to use this:

https://github.com/sot/fot_matlab_tools/blob/0fd5ab5f0afb019c06df4a979f247eb8d7f697bf/MissionScheduling/stars/StarSelector/%40StarSelector/private/getProsecoCatalog.m#L59

## Testing

- [N/A] Functional testing from Matlab Tools: functional testing deferred to the next MATLAB release.

Closes #202 
Closes #275

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -147,21 +147,6 @@ attributes::
   37879960 188.579307 2.444460  -567.34  -632.27  119.53 ...   10.70 0.000     1     1    25  6x6
   37882072 188.584100 1.455829  2197.62  1608.89 -436.73 ...   11.66 0.000     1     1    25  6x6
 
-The ``aca`` object provides a ``thumbs_up`` attribute which is
-a rough indicator that the catalog will likely pass or fail ACA review.
-The value of this is ``0`` for NOT OK and ``1`` for OK.  Note in
-this case that too few guide stars have been selected::
-
-  >>> aca.thumbs_up
-  0
-
-  >>> aca.acqs.thumbs_up
-  1
-  >>> aca.guides.thumbs_up
-  0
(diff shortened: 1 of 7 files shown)


jeanconn:
It looks like there is also some thumbs_up code in sparkles, but it reads to me that it would just not do anything if these properties didn't exist.  The sparkles unit tests should probably also run against this change.

jskrist:
There is a reference to the `thumbs_up` property in the python code our MATLAB tools call, but our MATLAB code doesn't actually use this data.

We will need to make sure that the three lines that reference the `thumbs_up` property are deleted in the same MATLAB Tools release (or before) that we start using this change.

taldcroft:
> We will need to make sure that the three lines that reference the thumbs_up property are deleted

Probably a good idea to open a Jira ticket on this to make sure it gets tracked.

jskrist:
Great idea.
- [X] Add JIRA ticket to track the necessary MATLAB change.

PR #334, opened 2020-07-31
Set to use 3.8 in flake8 check

jeanconn:
## Description

Set to use 3.8 in github workflow flake8 check

## Testing

- [ ] Passes unit tests on MacOS, linux, Windows (at least one required)
- [ ] Functional testing

Fixes #

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -9,10 +9,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
-    - name: Set up Python 3.6
+    - name: Set up Python 3.8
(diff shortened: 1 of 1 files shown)


.github/workflows/flake8.yml
```diff
@@ -9,10 +9,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
-    - name: Set up Python 3.6
+    - name: Set up Python 3.8
       uses: actions/setup-python@v1
       with:
-        python-version: 3.6
+        python-version: 3.8
```
jeanconn, line 15: I don't know if this is enough to "just work" but...

PR #340, opened 2020-09-23
Add include/exclude_ids to allowed fid arg test

jeanconn:
## Description

Add include/exclude_ids to allowed fid arg test

## Testing

- [x] Passes unit tests on linux (not sure how I missed this before)


jeanconn:
I guess I missed this because of the interaction between #331 and #332

jeanconn:
And I guess I needed to put test_allowed_kwargs in my description because @taldcroft  just fixed this again in #343.

PR #344, opened 2020-09-28
Add img_size guide parameter to control guide star image readout size

### LINKED ISSUE #337: Allow more control of guide readout size
It looks like the readout size is presently inferred from the number of fids

https://github.com/sot/proseco/blob/ea711053cdf83553d471be93dac28a7e32d53493/proseco/catalog.py#L442

For the activity to use some 6x6 readout windows on some ERs, should we add an option to override the size for each guide star?  Would something equivalent to include_halfws_acq work? Or would we want to override size (SZ?) it on a per-obsid basis?  I suppose I am a fan of getting *something* in here, even if it is just for this data collection project, as it will be good to see the right catalogs in the proseco pkl.

jeanconn:
## Description

Add img_size parameter to control box size

## Testing

- [x] Passes unit tests on Linux
- [x] Functional testing (test added)

Fixes #337 

Requires:
 - [ ] API test with Matlab tools

### DIFF
--- a/proseco/catalog.py
+++ b/proseco/catalog.py
@@ -62,8 +62,8 @@ def get_aca_catalog(obsid=0, **kwargs):
     :param exclude_ids_fid: list of fiducial lights to exclude by index
     :param include_ids_guide: list of AGASC IDs of stars to include in guide catalog
     :param exclude_ids_guide: list of AGASC IDs of stars to exclude from guide catalog
-    :param star_img_sz: readout window size for stars (6, 8, or None).
-                        With a value of ``None`` infer (8 for no fids, 6 for fids).
+    :param img_size_guide: readout window size for guide stars (6, 8, or ``None``).
+                           For default value ``None`` use 8 for no fids, 6 for fids.
     :param optimize: optimize star catalog after initial selection (default=True)
     :param verbose: provide extra logging info (mostly calc_p_safe) (default=False)
     :param print_log: print the run log to stdout (default=False)
@@ -104,6 +104,7 @@ def _get_aca_catalog(**kwargs):
     assembling the merged aca catalog.
     """
     raise_exc = kwargs.pop('raise_exc')
+    img_size_guide = kwargs.pop('img_size_guide', None)
 
     aca = ACATable()
     aca.call_args = kwargs.copy()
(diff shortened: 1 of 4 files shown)


proseco/catalog.py
```diff
@@ -147,7 +149,7 @@ def _get_aca_catalog(**kwargs):
     # impacting operational work (call from Matlab).
     try:
         aca.log('Starting merge_cats')
-        merge_cat = merge_cats(fids=aca.fids, guides=aca.guides, acqs=aca.acqs)
+        merge_cat = merge_cats(fids=aca.fids, guides=aca.guides, acqs=aca.acqs, **kwargs)
```
jeanconn, line 152: Not sure if passing all the kwargs to merge_cats makes sense.

proseco/catalog.py
```diff
@@ -439,7 +441,10 @@ def merge_cats(fids=None, guides=None, acqs=None):
         fids['p_acq'] = 0
         fids['sz'] = '8x8'
 
-    guide_size = '8x8' if len(fids) == 0 else '6x6'
+    if 'star_img_size' in kwargs and kwargs['star_img_size'] is not None:
```
jeanconn, line 444: This is the way my brain is working today but this could probably be trimmed down.

taldcroft:
@jeanconn - I had a number of suggestions and in the end it was easier to just do them in code. I think this is ready so now you should review.

taldcroft:
@jskrist - just want to confirm for MATLAB API that you have the machinery to include this optional parameter as necessary.

 We are imagining (like the include/exclude interface) that there is a text option box to manually set the guide star image readout size. It would normally be blank in which case ORviewer does not pass that parameter into the `get_aca_catalog` call and the Python default of `None` is used.  But if a value is entered then that gets passed as an integer. Legal values are 6 or 8.

Writing this down we should probably allow 4, which is never used in operational practice but it is a legal value.

proseco/guide.py
```diff
@@ -69,8 +69,8 @@ def get_guide_catalog(obsid=0, **kwargs):
 
 class ImgSizeMetaAttribute(MetaAttribute):
     def __set__(self, instance, value):
-        if value not in (6, 8, None):
-            raise ValueError('img_size must be 6, 8, or None')
+        if value not in (4, 6, 8, None):
+            raise ValueError('img_size must be 4, 6, 8, or None')
```
jeanconn, line 73: Looks good.  I was thinking this would get checked on the sparkles side but fine to do it here.

jeanconn (commented):
These changes look good to me.  I like the idea of moving the parameter to be an attribute only on the guide catalog.  Seems fine to add 4x4.  And thanks for cleaning up the test by adding an empty dark map and token catalog to speed it up.  I note I can't approve my own PR or apparently change owner, so this is just a comment.

taldcroft (approved):

jeanconn:
I note I just asked the question in https://github.com/sot/sparkles/issues/147 .. should we set the OR default size somewhere in the proseco config?

taldcroft:
> should we set the OR default size somewhere in the proseco config?

This is now coded in the `GuideTable.get_img_size()` method for any applications to use.  Does that work?

PR #345, opened 2020-11-09
Use AGASC supplement in proseco

taldcroft:
## Description

Use the AGASC supplement in proseco. This behavior is always enabled except through use of a decorator for testing.

This requires https://github.com/sot/agasc/pull/56.

## Testing

- [x] Functional testing (new unit test)


jeanconn:
Thanks Tom!  In our meeting I was wondering if it would be useful to test out the new agasc API by using it this way.

proseco/tests/test_catalog.py
```diff
@@ -46,7 +47,24 @@ def test_get_aca_catalog_49531():
     assert len(aca.fids) == 0
 
 
-@pytest.mark.skipif('not HAS_SC_ARCHIVE', reason='Test requires starcheck archive')
+@pytest.mark.skipif(not HAS_SC_ARCHIVE, reason='Test requires starcheck archive')
+def test_get_aca_catalog_20603_with_supplement():
```
jeanconn, line 51: This test looks to assume mag updates for a specific obsid.  I was just thinking that one possible path to promotion might be to promote all the agasc code and then promote the supplement separately (which I think would have this test failing).  If we are planning on promoting the supplement before the code or together I suppose this is fine.

proseco/characteristics.py
```diff
@@ -42,56 +41,11 @@
 bad_pixels = [[-245, 0, 454, 454],  # Bad column
               [-374, -374, 347, 347]]  # Fid trap
 
-bad_star_set = set([36178592,
```
jeanconn, line 45: And I *think* we had these in here before we started syncing the supplement (and as a resource for testing) so these can all go... but I don't remember if there was any other reason for the hardcoded list.

jeanconn (approved):

proseco/tests/test_catalog.py
taldcroft: Fixed

PR #346, opened 2020-11-19
Support monitor windows and full catalog

### LINKED ISSUE #305: Add monitor star handling to proseco (new `mon_star` argument)
Add a new keyword argument to the `get_aca_catalog` function in order to support proseco handling of monitor window requests in the OR `STAR=(RA, Dec, Mag)` or scheduling of monitor windows in ERs.

Proseco will then handle all aspects of deciding what to do with the request, following the defined process below for ORs (for `function=0` defined below).

If it is not initially selected, try selecting a catalog with the monitor star forced for inclusion as a guide star:
- If no new critical messages from sparkles appear relative to the initially-selected catalog, convert to guide (with 8x8 readout in slot 7)
- Else schedule the star as a monitor star.

This requires:
- New argument passed by ORviewer
- ORviewer needs to take the image readout size (6x6, 8x8) from the proseco catalog.
- Formal integration of sparkles into the star selection process.  This is a conceptual barrier rather than technical, but not a problem.

See also https://occ-cfa.cfa.harvard.edu/browse/MATLAB-11472

## Proseco API

One keyword argument `monitors` which can accept a list of lists (or 2-d array of floats) in the form:
```
[
 [coord0, coord1, coord_type, mag, function],  # Monitor window 1 (req'd)
 [coord0, coord1, coord_type, mag, function],  # Monitor windows 2, 3, ... N (optional)
 ..
]
```
If passed as a list of lists, the input will be converted to a numpy 2-d array (`np.array(monitors)`) which will have shape `(N, 5)` where `N` is the number of monitor windows.

- `coord0`: first coordinate (RA, row, or yag)
- `coord1`: second coordinate (Dec, col, or zag)
- `coord_type`: 0: RA/Dec (deg), 1: row/col (pixel), 2: yag/zag (arcsec)
- `mag`: star magnitude (only used for commanded MAXMAG for monitor window requests)
- `function`: 
  - 0: Schedule as a guide star if it meets requirements and does not introduce critical warnings. Always using brightest star as designated track star.
  - 1: Schedule as a guide star if it is in the AGASC (raises an exception if not).
  - 2: Schedule as monitor window that tracks the brightest star
  - 3: Schedule as monitor window that is fixed (no tracking)

Use cases:
- OR-specified monitor target `function=0`
  - This is the most common configuration and let's proseco decide how to schedule the OR
  - Coordinates provided as RA, Dec
- OR-specified monitor target `function=1`
  - Force scheduling as a guide star in a case where proseco would normally schedule as a monitor window.
  - E.g. a star where the observer provides information on mag or variability that is different from the AGASC (it is currently brighter than in the AGASC)
- OR-specified monitor target `function=2`
  - Force scheduling as a monitor window in a case where proseco would normally schedule as a guide star.
  - E.g. a star where the observer provides information on mag or variability that is different from the AGASC (fainter or more variable).
- Fixed-location ER monitor window `function=3`
  - Used for engineering investigation, e.g. looking at a specific flickering pixel
  - `coord0,1` as yag/zag or row/col, `coord_type=1 or 2`
  - Most commonly row/col, so we could just drop `coord_type=2` for simplicity.

taldcroft:
## Description

This is to address #305 and make the `aca` catalog be the complete specification of the star catalog that ORviewer will simply translate into a DOT star catalog.

## Testing

- [x] Passes unit tests on MacOS (with many new unit tests)
- [x] Functional testing

## Functional testing

- [Validate monitor star catalogs for 2020 and 2021](https://nbviewer.jupyter.org/url/cxc.cfa.harvard.edu/mta/ASPECT/proseco/validate-mon-catalogs.ipynb)
- [Validate full catalog generation for catalogs in 2021](https://nbviewer.jupyter.org/url/cxc.cfa.harvard.edu/mta/ASPECT/proseco/validate-full-catalogs.ipynb)

```
from proseco.tests.test_common import mod_std_info

row = 10
col = 20
mag = 7.5
monitors = [[row, col, 1, mag, 2],  # Tracking MON
            [row + 350, col + 400, 1, mag, 3],  # Fixed MON
            [1053.38, -275.16, 2, 8.0, 1]  # Guide from MON (force 8x8)
            ]
aca = get_aca_catalog(monitors=monitors, **mod_std_info(n_guide=5, n_fid=1))

aca
<ACATable length=12>
    0     1         1  FID  8x8   0.000    7.00    8.00   922.59 -1737.89     1     1    25
    1     2    139192  BOT  6x6   0.981    7.77    9.27   587.27   802.49    28     1   160
    2     3 611190016  BOT  6x6   0.982    6.97    8.47   175.44 -1297.92    28     1   160
    5     4 611192384  GUI  8x8   0.000    8.19    9.69  1053.38  -275.16     1     1    25  <== Guide from MON
    6     5       100  MON  8x8   0.000    7.50    9.00   -19.39    75.30     2     0    25  <== Tracking MON
    7     6       201  MON  8x8   0.000    7.50    9.00 -1751.09  2061.24     7     0    25  <== Fixed MON
    3     7 688523960  ACQ  6x6   0.983    5.93    7.43  -202.71 -1008.91    28     1   160
    4     8 611192384  ACQ  6x6   0.979    8.19    9.69  1053.38  -275.16    28     1   160
    5     9 688521312  ACQ  6x6   0.976    8.45    9.95  -739.68 -1799.85    28     1   160
    6    10 611189488  ACQ  6x6   0.972    8.71   10.21  1536.83  -786.12    28     1   160
    7    11 688522008  ACQ  6x6   0.963    9.04   10.54  -743.78 -2100.94    28     1   160
    0    12    134680  ACQ  6x6   0.964    9.05   10.55  1314.00  1085.73    28     1   160
```

Fixes #305 

cc: @jeanconn @jskrist


proseco/guide.py
```diff
@@ -109,7 +134,154 @@ def reject(self, reject):
...
+
+                # Make a stub row for a MON entry using zero everywhere. This
+                # also works for str (giving '0').
+                mon = {col.name: col.dtype.type(0) for col in self.itercols()}
+                # These type codes get fixed later in merge_catalog
+                mon['type'] = 'MFX' if monitor['function'] == MonFunc.MON_FIXED else 'MTR'
+                mon['sz'] = '8x8'
+                mon['dim'] = -999  # Obviously bad value for DTS, gets fixed later.
+                mon['res'] = 0
+                mon['halfw'] = 20
```
jeanconn, line 263: I haven't dug up the manual but note that starcheck looks to have options to support 10 15 and 20 based on IMGSZ

https://github.com/sot/starcheck/blob/f98cd278591714b84d4c69542c235f61b633eb01/starcheck/src/lib/Ska/Starcheck/Obsid.pm#L763

jeanconn:
Is the stuff in `validate` valuable enough to check-in or does it live in another repo?  I'd like to be able to run the validation notebook and fiddle with stuff.

proseco/guide.py
```diff
@@ -49,21 +56,39 @@ def get_guide_catalog(obsid=0, **kwargs):
...
+    # Process monitor window requests, converting them into fake stars that
+    # are added to the include_ids list.
+    guides.process_monitors_pre()
+
     # Do a first cut of the stars to get a set of reasonable candidates
     guides.cand_guides = guides.get_initial_guide_candidates()
 
+    # Process guide-from-monitor requests by finding corresponding star in
+    # cand_guides and adding to the include_ids list.
+    # guides.process_monitors_pre2()
```
jeanconn, line 68: ?

taldcroft:
> Is the stuff in validate valuable enough to check-in

Can you be more specific about the "stuff"?  The two notebooks mentioned in the description are intended as a permanent "published" version of the validation. This follows the precedent of the other validation notebooks that were presented for proseco promotion. You can download those two notebooks locally and run them and fiddle.

proseco/guide.py
taldcroft: Oops, that's a leftover. The previous comment mentioning fake stars is also a leftover. Will fix.

proseco/guide.py
```diff
@@ -109,7 +134,154 @@ def reject(self, reject):
     include_ids = AliasAttribute()
     exclude_ids = AliasAttribute()
 
-    def get_img_size(self, n_fids):
+    def process_monitors_pre(self):
+        """Process monitor window requests, converting them into fake stars that
```
taldcroft, line 138: Fix this docstring.

jeanconn:
Regarding `validate`, my mistake.  I didn't see the routines in this PR, was working in another directory, had /home/jeanconn/git/proseco in my PYTHONPATH, and forgot that `validate/starcheck_proseco.py` was from previous validation and also wouldn't import from my PYTHONPATH that way.  I suppose I could add /home/jeanconn/git/proseco/validate to my PYTHONPATH to have access if I want to import starcheck_proseco.py without being local or some such.

taldcroft:
@jeanconn - you have a point about that particular function `get_starcheck_obs_kwargs` being of use in a broader context. So I moved it into `tests.test_common` in e5305de.  I updated the two validation notebooks accordingly and also took out the `/Users/aldcroft` bit (but they still assume `~/git/agasc` exists and points to something useful).

jeanconn:
In my initial review, interface and testing/validation looks great.  

I'm still a little confused about some of the internals.  The redefinition of n_guide as guide + mon has tripped me up a few times (maybe there is another path to be taken there?), as has the fact that plain monitor stars are in cat.guides (I would have expected cat.mons or some such).

taldcroft:
> Oh, and if MON stars stay in `guides` I think we'll need fixes in other pieces like the guide star reporting, and maybe https://github.com/sot/sparkles/blob/37a56f5bedf7263da777a3ccbffdf581d212eaae/sparkles/core.py#L471 ?

Good point. I'll look into changing this.

taldcroft:
About `n_guide`, I'm not sure there is an easy solution because of the auto-conversion capability. When you call `get_aca_catalog` you won't know  exactly how many guide stars you'll get. The only thing you know is how many slots are allocated for guide + mon. We could rename that to `n_guide_mon`, but I don't think that is worth the churn. I can certainly update the docs at least to reflect the updated meaning of `n_guide`.

Also, it may help that with splitting out monitors to a new `mons` table, the attribute `n_guide` will correctly reflect the number of guide stars selected. It just won't necessarily match the input `n_guide` arg, but that was always the case. So maybe all good?

jeanconn:
Right, I totally understand no easy solution on n_guide as a call_arg.  I was initially confused, for example, that even if I specified the MON to *not* be auto that I still had to set n_guide to be "desired slots for guides + mons" or "most likely 8 - n_fids".  But I see that's probably a better convention than having n_guide as a call arg behave differently for auto MON -> guide observations.

So yes, I think if the attribute n_guide matches the number of selected guide stars we should be a little more consistent even if n_guide ends up a bit overloaded.

jeanconn:
Regarding external API, I don't entirely recall how we tried to handle catalog slot/idx ordering oddities for a BOT star used as a MON when there is also a GUI in the catalog.  For example in this constructed example:

```
obsid = 23378
acas = pickle.load(gzip.open("/data/mpcrit1/mplogs/2020/NOV2320/oflsa/output/NOV2\
320A_proseco.pkl.gz"))
aca = acas[obsid]
call_args = aca.call_args.copy()
call_args['n_guide'] = 5
star = (197.029792,-65.306028,5.530000,'MON')
monitors = [[star[0], star[1], MonCoord.RADEC, star[2], MonFunc.GUIDE]]
call_args['monitors'] = monitors
cat = proseco.get_aca_catalog(**call_args)
acar = cat.get_review_table()
acar.run_aca_review()
```

```
<ACATable length=12>
 slot  idx      id     type  sz   p_acq    mag    maxmag   yang     zang    dim   res  halfw
int64 int64   int64    str3 str3 float64 float64 float64 float64  float64  int64 int64 int64
----- ----- ---------- ---- ---- ------- ------- ------- -------- -------- ----- ----- -----
    0     1          2  FID  8x8   0.000    7.00    8.00  -773.20 -1742.03     1     1    25
    1     2          3  FID  8x8   0.000    7.00    8.00    40.01 -1871.10     1     1    25
    2     3          4  FID  8x8   0.000    7.00    8.00  2140.23   166.63     1     1    25
    3     4 1179272040  BOT  6x6   0.972    7.88    9.38  2389.67    86.64    28     1   160
    4     5 1179272560  BOT  6x6   0.977    8.39    9.89  2178.81 -1608.38    28     1   160
    5     6 1179272760  BOT  6x6   0.976    8.49    9.99  1076.66    41.61    28     1   160
    7     7 1179273480  BOT  8x8   0.983    6.12    7.62    69.41    32.79    28     1   160
    6     8 1179403384  GUI  6x6   0.000    8.41    9.91  1452.47 -1660.32     1     1    25
    0     9 1179913456  ACQ  6x6   0.981    7.88    9.38 -1966.52  -597.39    28     1   160
    1    10 1179269944  ACQ  6x6   0.974    8.60   10.10  -339.56  1053.65    28     1   160
    2    11 1179388832  ACQ  6x6   0.973    8.65   10.15  -446.24 -2302.05    28     1   160
    6    12 1179271096  ACQ  6x6   0.969    8.86   10.36  -634.59   496.25    28     1   160
```

I don't know if the only reason for the previous ordering was to keep some SAUSAGE code happy or if there might be a reason that slots need to show up in a certain order in the actual command (if that is the case, it seems like we'd need to split the slot 7 BOT into a GUI and an ACQ?).  We talked about ordering a bit in the aca@ thread "Change in ACA catalog ordering (non-impacting, hopefully)", but I think in the end it is just a matter of testing?

taldcroft:
@jeanconn - for better or worse, your comments inspired me to totally refactor how the monitor window handling is done. Monitors now look a lot more like acqs, guides, and fids. It is much better I think, but ended up taking most of the day just to get back to passing the same tests.  :smile:

taldcroft:
>  but I think in the end it is just a matter of testing?

Yup. We'll test and in the unlikely event that something in MATLAB breaks we will revisit (mostly likely by fixing MATLAB, not changing how we do operations). In the end Eric was OK with the changes from the OBC / FOT operations perspective.

jeanconn:
Thanks for the reorganization to make `.mons` !  What were your thoughts about disposition for `n_guide`?

jeanconn:
I'll have to dig in to the code a bit more; I just realized I'm not clear on which position is being used when the monitor target resolves to a useable guide star.  For within 2 arcsecs it doesn't really matter but I confess to not being sure what is proposed.

taldcroft:
> I just realized I'm not clear on which position is being used when the monitor target resolves to a useable guide star. For within 2 arcsecs it doesn't really matter but I confess to not being sure what is proposed.

If the monitor request gets converted to a guide star, then this means the resolved AGASC ID is force-included as a guide star, with the exception that it ends up with 8x8 in all cases. The position and mag from the monitor target in the OR list is ignored (observers are not trusted with guide stars).

jeanconn:
And for the case when an agasc star is identified (and put in the proseco table) but it is left as a MON, the position of the agasc star is ignored?

taldcroft:
> And for the case when an agasc star is identified (and put in the proseco table) but it is left as a MON, the position of the agasc star is ignored?

Correct. It uses the AGASC `MAG_ACA` but the observer's position. The mag is basically informational, but the rationale is that the user-supplied mag is not in any well-defined mag system, and we we expect in that star catalog column is the best estimate of ACA mag.

jeanconn (approved):
I think the work and changes that went into addressing my last set of questions makes this PR complete.  I have remaining concerns about perhaps needing still more testing/review for the catalog order / convention of commanded slot order changing in the final commanded catalog.

PR #349, opened 2020-11-27
Built-in function for diffing ACA catalogs

taldcroft:
## Description

Seems like this wheel gets re-invented a lot. I need this for the monitor window / full catalog validation so I decided to put this in as a new module in proseco.

## Testing

- [x] No unit tests
- [x] Functional testing

I'm using this for the #346. For catalogs with a monitor request in 2020 it gives a result like this. (Flight catalogs on the left, new catalogs on the right):  https://cxc.harvard.edu/mta/ASPECT/tmp/flight-aca-diff.html

EDIT: that HTML output is outdated.  To see this in action try in Jupyter.
```
from proseco.tests.test_common import mod_std_info
from proseco.diff import catalog_diff

aca0 = get_aca_catalog(**mod_std_info(n_guide=5, n_fid=3),
                       include_ids_guide=[611192064])
aca1 = get_aca_catalog(**mod_std_info(n_guide=6, n_fid=2))
names = 'id type sz dim res halfw'
catalog_diff([aca0, aca1], [aca1, aca0], labels=['Obsid 1', 'Obsid 2'], style='html', names=names)
```

Note that in #346 I've given up on maintaining the same slot / index order. The existing order which is by mag, except for BOT stars where it is by AGASC ID, is just silly (and turned out to be rather difficult to replicate with MON stars in play). The new order is by mag within each category.


taldcroft:
Oops, need to pull out just the last commit for this.

taldcroft:
I've tested this pretty thoroughly so merging now. If there are any changes that would come from review they can be done as a separate PR.

PR #354, opened 2021-03-18
WIP: Save agasc supplement info in meta

jeanconn:
## Description

Save the agasc supplement info in the meta for a catalog if the supplement data was used.

## Testing

- [x] Functional testing - new simple asserts added to existing test

### DIFF
--- a/proseco/core.py
+++ b/proseco/core.py
@@ -17,6 +17,8 @@ from chandra_aca.transform import (yagzag_to_pixels, pixels_to_yagzag,
                                    radec_to_yagzag, yagzag_to_radec)
 from chandra_aca.aca_image import AcaPsfLibrary
 import agasc
+import agasc.agasc
+from agasc import get_supplement_table
 from Quaternion import Quat
 from mica.archive.aca_dark import get_dark_cal_image, get_dark_cal_id
 
@@ -521,6 +523,8 @@ class ACACatalogTable(BaseCatalogTable):
     t_ccd_guide = MetaAttribute()
     date = MetaAttribute()
     dark_date = MetaAttribute()
+    supp_agasc_version = MetaAttribute()
+    supp_last_updated = MetaAttribute()
     dither_acq = MetaAttribute()
     dither_guide = MetaAttribute()
     detector = MetaAttribute()
@@ -669,6 +673,7 @@ class ACACatalogTable(BaseCatalogTable):
                 stars = StarsTable.from_agasc(self.att, date=self.date, logger=self.log)
             else:
                 stars = StarsTable.from_stars(self.att, self.stars, logger=self.log)
+            stars.set_agasc_supp_version()
         else:
             stars = acqs.stars
 
@@ -982,6 +987,8 @@ class StarsTable(BaseCatalogTable):
 
     """
     att = QuatMetaAttribute()
+    supp_agasc_version = MetaAttribute()
+    supp_last_updated = MetaAttribute()
 
     # StarsTable attributes, gets set in MetaAttribute or AliasAttribute
     allowed_kwargs = set()
@@ -995,6 +1002,28 @@ class StarsTable(BaseCatalogTable):
                 pass
 
             return null_logger
+
+    def set_agasc_supp_version(self):
+        """
+        Get the versions of the agasc supplement used for the star data for catalog
+        determination and save in the appropriate MetaAttributes.
+        """
+
+        # It seems incorrect to save the supplement versions that were used to
+        # update stars if the supplement was not actually enabled/used.
+        # This block copies the logic from agasc.update_from_supplement
+        # (which perhaps should be generalized in the interface).
+        supplement_enabled_env = os.environ.get(agasc.SUPPLEMENT_ENABLED_ENV,
+                                                agasc.agasc.SUPPLEMENT_ENABLED_DEFAULT)
+        if supplement_enabled_env not in ('True', 'False'):
+            raise ValueError(
+                f'{agasc.SUPPLEMENT_ENABLED_ENV} env var must be either "True" or "False" '
+                f'got {supplement_enabled_env}')
+        if supplement_enabled_env == 'True':
+            agasc_versions = get_supplement_table('agasc_versions')
+            self.supp_agasc_version = agasc_versions['supplement'][0]
+            last_updated = get_supplement_table('last_updated')
+            self.supp_last_updated = last_updated['supplement'][0]
 
     def get_catalog_for_plot(self):
         """Return value of ``catalog`` for plot_stars() call for this object
@@ -1157,6 +1186,7 @@ class StarsTable(BaseCatalogTable):
         :param agasc_id: AGASC ID of the star to add
         """
         stars = StarsTable.from_agasc_ids(self.meta['q_att'], [agasc_id])
+        stars.set_agasc_supp_version()
         self.add_row(stars[0])
 
     def add_fake_constellation(self, n_stars=8, size=1500, mag=7.0, **attrs):
--- a/proseco/tests/test_catalog.py
+++ b/proseco/tests/test_catalog.py
@@ -2,6 +2,7 @@
 import copy
 import os
 
+from distutils.version import LooseVersion
 import matplotlib
 from Quaternion import Quat
 
@@ -11,6 +12,7 @@ import pickle
 import pytest
 from pathlib import Path
 
+from cxotime import CxoTime
 import numpy as np
 import mica.starcheck
 import agasc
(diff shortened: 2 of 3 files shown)


proseco/core.py
```diff
@@ -996,6 +1003,28 @@ def null_logger(*args, **kwargs):
...
+        # It seems incorrect to save the supplement versions that were used to
+        # update stars if the supplement was not actually enabled/used.
+        # This block copies the logic from agasc.update_from_supplement
+        # (which perhaps should be generalized in the interface).
+        supplement_enabled_env = os.environ.get(agasc.SUPPLEMENT_ENABLED_ENV,
+                                                agasc.agasc.SUPPLEMENT_ENABLED_DEFAULT)
+        if supplement_enabled_env not in ('True', 'False'):
+            raise ValueError(
+                f'{agasc.SUPPLEMENT_ENABLED_ENV} env var must be either "True" or "False" '
+                f'got {supplement_enabled_env}')
```
jeanconn, line 1021: I didn't see a sensible way around this which suggests to me that it might be helpful to add something to the agasc module to handle the logic.  But I realize that I'm the only one interested in this.

proseco/tests/test_catalog.py
```diff
@@ -2,6 +2,7 @@
 import copy
 import os
 
+from distutils.version import LooseVersion
```
jeanconn, line 5: I don't recall if this is still the recommended way to compare versions.

proseco/tests/test_catalog.py
```diff
@@ -69,6 +71,11 @@ def test_get_aca_catalog_20603_with_supplement():
     assert (len(aca_no.acqs) != len(aca.acqs)
             or np.any(aca_no.acqs['mag'] != aca.acqs['mag']))
 
+    assert LooseVersion(aca.supp_agasc_version) > LooseVersion('4.10')
+    assert CxoTime(aca.supp_last_updated) > CxoTime('2021-03-01')
+
+    assert aca_no.supp_agasc_version is None
```
jeanconn, line 77: For tests I didn't have much in mind, so just threw in checks that the supplement if used was at least as new (with margin) from the one I'm seeing in the 'rc' area.

jeanconn:
For this, I didn't add anything fancy, like the checks we have for `dark` that the date that corresponds the data in the structure is the same as the one that made the catalog ...

proseco/core.py
javierggt: I agree that nothing out of agasc should be doing these kind of manipulations to interpret the environment variables.

Also, I am assuming it would not be possible to fool this determination by using the `set_supplement_enabled` context manager. 

It also bugs me that even though this is a boolean flag, it does a string comparison. I mean, it bugs me to read:
```
if supplement_enabled_env == 'True':
```
instead of 
```
if supplement_enabled_env:
```

jeanconn: I don't know if I've thought of the ways somebody might fool this determination using `set_supplement_enabled`.  I think `agasc` probably needs a `get_supplement_enabled` state checker to isolate scope to that module.  *But* something like this could work as a `proseco` update without more changes in the `agasc` module...

jeanconn: With regard to it being a string comparision, it sounds to me like the string comparison isn't the issue... you are just bothered that environment variables don't have even-weak types?

javierggt: I would first interpret the env variables to produce a bool, and then use the bool in the if statement. In this case it is the same, because the if statement follows right away, so it is just a matter of style.

If you merge this change, and then if you ever modify how the env variables are interpreted inside agasc, you might get an inconsistency. I know I would forget about this. This is the main issue (and you original comment). That's why it might be a good idea to do it right.

jeanconn: Right, so I agree that I thought this proseco code was awkward, but I figured I'd write it out as a bit of a use case so we could define the best changes in `agasc`.  I would figure we'd pull out the block in update supplement https://github.com/sot/agasc/blob/dddfab67a67b783c97af358fd14e4fbbf1fcd8cd/agasc/agasc.py#L503 into a new method and export it.

javierggt: Yes. It is a bit more than that:
- First thing is to move the code in [agasc.py#L504](https://github.com/sot/agasc/blob/master/agasc/agasc.py#L504) into its own function.
- Then one needs to add this function to `agasc. __all__`
- and remove `'SUPPLEMENT_ENABLED_ENV'`. The point here is that SUPPLEMENT_ENABLED_ENV is not the right thing to expose, because the right thing is a combination of SUPPLEMENT_ENABLED_ENV and SUPPLEMENT_ENABLED_DEFAULT.

Since this was explicitly named in `agasc.__all__`, then the question is if it is in use in other places already.

jeanconn: Yes, that's exactly what I meant in the previous comment.

taldcroft:
How might one actually use the value of `supp_agasc_version`?  I can see how the `supp_last_updated` can (in theory) be used to recover a particular version, but I don't have the imagination to see how the version of the `agasc` package which was used to update the supplement that time is actionable.

jeanconn:
I thought it could be useful for the same reason that we put it in the file (which you may also disagree with).  For now, for example, we've got a testing file built with a version of `agasc` before it has been officially promoted.  With an eye to real future improvements / development of the code that makes the data, it would be useful to track the version of the `agasc` module that made the code.  So I just figured date + version was really what made the data file unique.

But could also cut it for simplicity.

proseco/core.py
taldcroft: > and remove 'SUPPLEMENT_ENABLED_ENV'

You can't remove this env var definition because this is the programmatic pathway to disabling use of the supplement for an entire module.  See e.g.: 
https://github.com/sot/proseco/blob/20dcd52b2d8f57f80752fd00d8ce06656f65d9f3/proseco/tests/test_catalog.py#L28

The API here is strict, namely the value of that environment variable (which is by definition a string) can have one of two values: `"False"` or `"True"`. Anything else is an exception and one gets a clear message of what went wrong and how to fix it. 

With that API defined it really is not a bad thing that string comparisons are used given the very simple logic that is contained in a few lines. Sure, we could import `ast` and do an `ast.literal_eval()` of the variable, but that hasn't really made things easier to understand IMO.

Moving the test logic (the line 504 stuff) into a separate function that is also exposed in the public API is a good idea.

jeanconn: I was thinking two things:

1) would remove SUPPLEMENT_ENABLED_ENV from proseco because that would be encapsulated in a new public method.

2) I was thinking that the programmatic pathway to disability use of the supplement in Python could just use another public method that encapsulated it, but maybe I have not thought that through.

javierggt:
> How might one actually use the value of supp_agasc_version?

I am also not sure we will use it, but supp_agasc_version tells you the version that was actually used to compute the magnitude, so if you ever wanted to compare two catalogs which differ because the star magnitudes were slightly different, you could trace the origin of the discrepancy.

proseco/core.py
```diff
@@ -996,6 +1003,28 @@ def null_logger(*args, **kwargs):
...
+        if supplement_enabled_env == 'True':
+            agasc_versions = get_supplement_table('agasc_versions')
+            self.supp_agasc_version = agasc_versions['supplement'][0]
+            last_updated = get_supplement_table('last_updated')
+            self.supp_last_updated = last_updated['supplement'][0]
```
taldcroft, line 1026: It's worth a comment here that `last_updated` is a 1-row table, and that `supplement` is the column that provides the time of last update of any table of the supplement (besides the version tables themselves).

proseco/core.py
```diff
@@ -996,6 +1003,28 @@ def null_logger(*args, **kwargs):
 
             return null_logger
 
+    def set_agasc_supp_version(self):
+        """
+        Get the versions of the agasc supplement used for the star data for catalog
```
taldcroft, line 1008: This doc string is not quite reflective of what the method is actually doing. There is one version and one last update time.

proseco/core.py
```diff
@@ -521,6 +523,8 @@ class ACACatalogTable(BaseCatalogTable):
     t_ccd_guide = MetaAttribute()
     date = MetaAttribute()
     dark_date = MetaAttribute()
+    supp_agasc_version = MetaAttribute()
+    supp_last_updated = MetaAttribute()
```
taldcroft, line 527: Use `supplement_last_updated`.

proseco/core.py
jeanconn: And do we want another convenience method in `agasc` for it or comment it here in `proseco`.

taldcroft:
> I am also not sure we will use it, but supp_agasc_version tells you the version that was actually used to compute the magnitude,

For a particular star mag update (one of the 80000 stars), you have no idea which version of `agasc` was used to compute that magnitude. The `supp_agasc_version` just tells you the installed version of `agasc` for the most recent update to the supplement file, but you have no `agasc` version that can be tied to a particular mag and mag error. You do of course have the star last update time, which can (with some effort) be tied to the version of `agasc`. That's what we might do in reality, but the `supp_agasc_version` being put in the pickle seems not useful.

taldcroft:
Not likely to ever get promoted.

PR #356, opened 2021-06-24
Get ACA penalty and planning limits from chandra_models

taldcroft:
## Description

This gets the ACA penalty and planning limits from chandra_models. It uses the module-level `__getattr__` to make these characteristics be lazily defined to prevent import-time issues.

The current code has been tested using a local version of chandra_models in the SKA data that has a tagged "release" from https://github.com/sot/chandra_models/pull/74 as the default model version. Ideally we should get a new version of chandra_models 3.35.1 promoted that includes only the ACA model update.

## Testing

- [ ] Functional testing

### DIFF
--- a/proseco/catalog.py
+++ b/proseco/catalog.py
@@ -44,6 +44,8 @@ def get_aca_catalog(obsid=0, **kwargs):
     :param man_angle: maneuver angle (deg)
     :param t_ccd_acq: ACA CCD temperature for acquisition (degC)
     :param t_ccd_guide: ACA CCD temperature for guide (degC)
+    :param t_ccd_penalty_limit: ACA CCD penalty limit for planning (degC). If not
+        provided this defaults to value from the ACA xija thermal model.
     :param date: date of acquisition (any DateTime-compatible format)
     :param dither_acq: acq dither size (2-element sequence (y, z), arcsec)
     :param dither_guide: guide dither size (2-element sequence (y, z), arcsec)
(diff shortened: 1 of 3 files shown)


jeanconn:
Regarding "The current code is using a local version of chandra_models that has a tagged "release" from sot/chandra_models#74 as the default model version.", do you just mean that in your testing you have a SKA set with a nonstandard version of chandra_models?  Or is there something in the code that I'm missing?

taldcroft:
> do you just mean that in your testing you have a SKA set with a nonstandard version of chandra_models? 

No, I changed my local "flight" Ska data version of chandra_models for testing purposes. Since (1) chandra_models is not currently used in any flight tools, and (2) the change is innocuous anyway, I was not concerned about this. My local mod will get wiped out the next time I do a ska_sync.

jeanconn:
I'm not sure how that is a "no"?

taldcroft:
I interpreted a "SKA set" to mean changing my SKA env var to a different value where the test Ska data were located.

jeanconn:
Gotcha.  Your SKA is set to a directory that has a nonstandard chandra_models was my point.  Though in your strategy you just used the one that you generally use.  Either way, there's nothing funny in the current PR/code to pick up a different version, it is just the test environment.

jeanconn (approved):
I like this implementation.  Though now that we have a good use case of get_xija_model_spec I'm wondering about the easiest way to specify a nonstandard location of the chandra_models repo for testing.  Env vars are annoying but maybe it is worth one?

PR #357, opened 2021-06-26
Add region of diminshed CCD response to bad_pixels

taldcroft:
## Description

Add a region of CCD pixels that appear to be associated with diminished response in the CCD. This is the proseco analog of the starcheck update https://github.com/sot/starcheck/pull/373.

## Testing

- [x] Passes unit tests on MacOS : one test failing right now.
- [x] Functional testing

### Functional testing
 
Ran star selection for obsids 48040 and 22957. Confirmed that with the patch in place, the stars which originally dithered over the bad region were not selected. Confirmed that without the patch those stars were selected as BOT.

### DIFF
--- a/proseco/characteristics.py
+++ b/proseco/characteristics.py
@@ -38,8 +38,13 @@ bad_pixel_dark_current = 700_000
 
 # Bad pixels.
 # Fid trap: http://cxc.cfa.harvard.edu/mta/ASPECT/aca_weird_pixels/
+# Diminished response: https://chandramission.slack.com/archives/G01LN40AXPG/p1624543385000700
+#   and subsequent discussion in nearby aspect-team threads.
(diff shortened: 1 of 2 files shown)


taldcroft:
There's a bit of a problem. The current implementation for bad pixels uses a convenient hack of making those pixels into super bright hot pixels. This works great for guide, but for acq it ends up spoiling a large region of the CCD which is actually just fine by making a bright imposter star. We need to figure out a better way.

jeanconn:
Yes.  I figured that setting these pixels to 700_000 might be inappropriate for ACQ but had not looked at the acq probability in this region to confirm it is "actually just fine".

jeanconn:
It would probably be most convenient to just to the exclusion for guide... but I suppose we don't want to park a fid light on these pixels either.

jeanconn (approved):

PR #359, opened 2021-06-28
Adding testing of guide bad pixel margin

taldcroft:
## Description

This started off as adding one pixel of margin to match starcheck and then adding a test, but when I did the testing it seems like there is already a bonus pixel in there somewhere. So this just adds testing.

## Testing

- [n/a] Functional testing

### DIFF
--- a/proseco/tests/test_catalog.py
+++ b/proseco/tests/test_catalog.py
@@ -487,7 +487,12 @@ def test_bad_pixel_dark_current():
     stars = StarsTable.empty()
     stars.add_fake_constellation(mag=np.linspace(8.0, 8.1, 4), n_stars=4)
     stars.add_fake_star(row=-205, col=450, mag=6.0, id=1)
+    # For acq (105 arcsecs away)
     stars.add_fake_star(row=-205, col=454 - 20 / 5 - 105 / 5, mag=6.0, id=2)
+
+    # For guide: 5 pixels + dither away (reject), 6 pixels + dither away (accept)
+    stars.add_fake_star(row=-150, col=454 - 5 - 20 / 5, mag=6.2, id=3)  # reject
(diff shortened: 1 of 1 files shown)


jeanconn:
I've completely forgotten why we use 4 as the pad for bad pixels and then there's another 1 in here

https://github.com/sot/proseco/blob/bff1f932f21e6cd0e6053a8d4fdd857528cdf364/proseco/guide.py#L1035

Probably just decided to keep tossing compromises at the problem...

taldcroft:
I was thinking that `+ 1` was about the width of the pixel itself. Assuming `cand['row']` is in 'edge' coordinates, then you need to add 1 to start counting 4 pixels from the top of the pixel.

proseco/tests/test_catalog.py
```diff
@@ -497,7 +502,7 @@ def test_bad_pixel_dark_current():
     assert np.all(aca.acqs.dark[-245 + 512:512, 454 + 512] == ACA.bad_pixel_dark_current)
 
     exp_ids = [2, 100, 101, 102, 103]
-    assert sorted(aca.guides['id']) == exp_ids
+    assert sorted(aca.guides['id']) == sorted(exp_ids + [4])
```
jeanconn, line 505: It might be more self-documenting in the test to just have exp_ids_acq and exp_ids_guide, but I think you prefer fewer variables.

taldcroft: My bar for code quality / readability is way lower in tests (both in writing and reviewing).

jeanconn: All good.  Just when doing the light code review here my process was ... "wait, why isn't id=4 in exp_ids?  Oh it probably isn't a good acq star (would guess it is too close to that bad pixel that is the point) and he uses exp_ids as the acq stars one line down".

jeanconn (approved):

PR #360, opened 2021-07-02
Remove default n_fid=3

taldcroft:
## Description

The issue with sparkles tests made me think about the default n_fid=3, which is not right in many cases. I think it is better to just force code to be explicit about this.

## Testing

- [x] Functional testing: checked with sparkles as well


jeanconn:
I agree the default was unhelpful.

jeanconn (approved):

PR #363, opened 2021-07-13
Set catalog n_guide appropriately in case of mon windows

taldcroft:
## Description

When getting a catalog via `aca = get_aca_catalog(...)`, the `aca.n_guide` attribute now reflects the input `n_guide` minus the number of scheduled monitor windows. This is effectively the number of requested guide stars, accounting for the possibility of monitor stars that are auto-converted from MON to GUI.

To clarify with an example, if `n_guide=5` and a `monitors` arg is provided with the position the target which happens to be a nice 8th mag star in the AGASC suitable for guiding, and the monitor function is `AUTO`, then `aca.n_guide` will be `5` and `aca.mons` will be an empty table. 

However, if that target is a 12th mag star which is not suitable for guiding, then `aca.n_guide` will be `4` and `aca.mons` will have one entry with a tracking monitor window.

## Testing

- [x] Passes unit tests on MacOS (with new tests)
- [n/a] Functional testing

### DIFF
--- a/proseco/catalog.py
+++ b/proseco/catalog.py
@@ -51,6 +51,16 @@ def get_aca_catalog(obsid=0, **kwargs):
     appending this string, e.g. with ``obs_text + '--force-catalog'`` in the
     call to ``get_aca_catalog``.
 
+    The input ``n_guide`` parameter represents the number of slots available for
+    the combination of guide stars and monitor windows (including both fixed and
+    tracking monitor windows). In most normal situations, ``n_guide`` is equal
+    to ``8 - n_fid``. The ``n_guide`` parameter is confusingly named but this is
+    because the actual number of guide stars is not known in advance in the case
+    of auto-conversion from a monitor request to a guide star. In actual
+    practice, what is normally known is how many slots are available for the
+    combination of guide stars and monitor windows, so this makes the call to
+    catalog creation simpler.
+
     NOTE on API:
 
     Keywords that have ``_acq`` and/or ``_guide`` suffixes are handled with
@@ -63,7 +73,7 @@ def get_aca_catalog(obsid=0, **kwargs):
     :param att: attitude (any object that can initialize Quat)
     :param n_acq: desired number of acquisition stars (default=8)
     :param n_fid: desired number of fid lights (req'd unless obsid spec'd)
-    :param n_guide: desired number of guide stars (req'd unless obsid spec'd)
+    :param n_guide: desired number of guide stars + monitor windows (req'd unless obsid spec'd)
     :param monitors: N x 5 float array specifying monitor windows
     :param man_angle: maneuver angle (deg)
     :param t_ccd_acq: ACA CCD temperature for acquisition (degC)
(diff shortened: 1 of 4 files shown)


proseco/catalog.py
```diff
@@ -199,6 +209,12 @@ def _get_aca_catalog(**kwargs):
     aca.guides = get_guide_catalog(stars=aca.acqs.stars, fids=aca.fids, mons=aca.mons,
                                    img_size=img_size_guide, **kwargs)
 
+    # Set output catalog aca.n_guide to reflect what the number of requested
+    # guide stars as determined in guide star selection processing. The input
+    # arg value of n_guide (which initially defines aca.n_guide) reflects the
+    # requested number of available slots for guide + monitor stars / windows.
+    aca.n_guide = aca.guides.n_guide
```
jeanconn, line 216: OK.  So I've gotten confused again.  From https://github.com/sot/proseco/pull/346#issuecomment-797572645 I thought the n_guide attribute was already going to be set to the number of selected guide stars as a side effect of the processing, but maybe I didn't completely validate.

proseco/catalog.py
```diff
@@ -199,6 +209,12 @@ def _get_aca_catalog(**kwargs):
     aca.guides = get_guide_catalog(stars=aca.acqs.stars, fids=aca.fids, mons=aca.mons,
                                    img_size=img_size_guide, **kwargs)
 
+    # Set output catalog aca.n_guide to reflect what the number of requested
+    # guide stars as determined in guide star selection processing. The input
```
jeanconn, line 213: I think maybe the "what" needs to be cut from this sentence for it to read as a sentence, but I also don't think that putting "requested" here helps (that seems to get easily confused with the n_guide call arg again as a kind of "request").

proseco/catalog.py
taldcroft: > Also, it may help that with splitting out monitors to a new mons table, the attribute n_guide will correctly reflect the number of guide stars selected. It just won't necessarily match the input n_guide arg

This PR is making that so. It *was* true previously for the the `aca.guides.n_guide` attribute, but not for `aca.n_guide`.

proseco/catalog.py
taldcroft: Done?

proseco/catalog.py
jeanconn: Thanks.  That's helpful for context.  So now from the comment I'm just confused about if aca.guides.n_guide is set to the number of guide stars requested of the guide algorithm or the number of guide stars selected by the guide algorithm.  I'll review.

jeanconn: OK.  I see for example that the sparkles code sets a local n_guide equal to the number of guide stars in the selected catalog and for its warning compares that to the attribute n_guide with, IIRC, the idea that is the helpful number of guide stars requested of the guide algorithm.  I think "number of requested guide stars as determined in guide star selection processing" is still a bit confusing but is a step in the right direction and we can come back to this if we come up with better language.

jskrist:
possible dumb question: What is the purpose of the "requested number of guide stars" being provided to the algorithm?  It seems to be overwritten based on what is actually found.  Is it there only for error checking at the end?

jeanconn:
Not a dumb question.  It has been a while but IIRC the number of guide stars requested "of the guide star selection algorithm" is there as an upper limit to help with trivial optimization.  If it is an OR and we only need at most 5 guide stars, the search stages and such will stop when we have 5 selectable stars.

jeanconn (approved):

PR #365, opened 2021-07-13
Copy back the original kwargs into the monitor catalog call args

### LINKED ISSUE #364: Issues with roll optimization of a catalog with monitor windows
The sparkles roll optimization code works from the call_args of the aca (ACACatalogTable) to run star selection and review on roll options.  It looks like this conflicts with the current proseco monitor code in a couple of ways.

1. It looks like in `_get_aca_catalog_monitors` the use of the "stub" catalog `monitors` as a reference causes inconsistencies

https://github.com/sot/proseco/blob/175eeeea561adff3561ccfd29e77fa64c5321dfa/proseco/catalog.py#L273

means that even for a case when the "try this as a guide star" case has failed, the kwargs are still set with a "run this as a guide star" in the returned aca_mon object at 

https://github.com/sot/proseco/blob/175eeeea561adff3561ccfd29e77fa64c5321dfa/proseco/catalog.py#L288

2. For either `aca_mon` or `aca_gui` returned from from `_get_aca_catalog_monitors`, the original monitors keyword call arg is lost.  Thus even if the value was "right" and function set to 2 for an aca_mon version of the catalog, would that be the most useful thing in the call args?  I think the actual original requested monitor type is the thing that, say, the sparkles roll optimization should have handy.

For a fix for both of these issues, I suggest that the `kwargs_orig` should just be copied back into the call args of the returned object.

For a test I was working with @jskrist 's

```
kw_dict = {'att': [ 0.32916333, -0.50759709,  0.07481427,  0.79271655],
 'date': '2021:116:16:42:09.065', 'detector': 'ACIS-I',
 'dither_acq': (7.9992, 7.9992), 'dither_guide': (7.9992, 7.9992),
 'man_angle': 86.49934496445843, 'n_acq': 8, 'n_fid': 3, 'n_guide': 5,
 'obsid': 23050.0, 'sim_offset': 0.0, 'focus_offset': 0.0,
 't_ccd_acq': -9.067548914167258, 't_ccd_guide': -8.616156814261098,
 't_ccd_penalty_limit': -6.8, 
 'monitors': [[335.516667,  58.675556,   0.      ,   9.63    ,   0.      ]]}
```

And running roll optimization and also seeing the BadMonitorError.

jeanconn:
## Description

Copy back the original kwargs into the monitor catalog call args

The ACACatalogTable object should have the original arguments and not any arguments that were used just to determine if the monitor window should be a MON or a guide-from-mon, and this change assures that.

## Testing

- [x] Functional testing - new test added in https://github.com/sot/sparkles/pull/156

Fixes #364

### DIFF
--- a/proseco/catalog.py
+++ b/proseco/catalog.py
@@ -274,6 +274,7 @@ def _get_aca_catalog_monitors(**kwargs):
 
     # Get the catalog and do a sparkles review
     aca_mon = _get_aca_catalog(**kwargs)
+    aca_mon.call_args = kwargs_orig.copy()  # Needed for roll optimization, see #364
     acar_mon = aca_mon.get_review_table()
     acar_mon.run_aca_review()
     crits_mon = set(msg['text'] for msg in (acar_mon.messages >= 'critical'))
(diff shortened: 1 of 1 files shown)


jeanconn:
I don't see any reason that the catalog needs to have the "as solved-for" monitor kwargs, but if there is another driver for the way this code was working, @taldcroft can speak to it.  I also figure he may want to go with a different fix strategy, but this fix can be tested to at least see if I have the scope correctly isolated.

taldcroft:
Looks good, but just move each of the lines setting the kwargs to be immediately after the line where the catalog is created. That makes the code localized. Also add a comment like `# Needed for roll optimization, see #364` in each case.

taldcroft (approved):
Approved, assuming this has been tested with the new sparkles test.

jeanconn:
I confirm this was re-tested with the new sparkles test.

PR #366, opened 2021-08-20
Improve handling for missing required args

taldcroft:
## Description

In the course of investigating the sparkles failure from https://github.com/sot/sparkles/pull/159, I noticed that the handling of missing required arguments could be improved.

This update catches exceptions in a likely place and provides a more useful error message.

## Testing

- [x] Passes unit tests on MacOS (with new tests)
- [x] Functional testing

### Functional testing

Ran the new exception cases in IPython and confirmed the expected exception chain for the obsid=99999 case.


taldcroft:
@jeanconn - awesome catch! The root cause is that the `required_attrs` was not set correctly for `ACATable`.

jeanconn (approved):
Nice!

PR #368, opened 2021-09-09
Move bright limits to approach or hit the new 5.2 mag onboard limit

jeanconn:
## Description

Move bright limits to approach or hit the new 5.2 mag onboard limit.
This shifts all of the effective bright limits in the code by 0.6 mags .

I note the grid acquisition model looks to spit out what-look-like-reasonable probabilities to mag = 5.0 .

## Testing


proseco/characteristics_guide.py
```diff
@@ -109,7 +109,7 @@
           {"Stage": 6,
            "SigErrMultiplier": 0,
            "ASPQ1Lim": 20,
-           "MagLimit": [5.6, 10.3],
+           "MagLimit": [5.2, 10.3],
```
jeanconn, line 112: For this one entry, I've only shifted it brighter by 0.4 mags instead of the 0.6 that I used for the rest.  But for the guide code, I think the candidate cut was previously at 5.8 anyway, so I don't know if the 5.6 here (brighter than that) in the stages was actually doing anything.  Will check.  The question is really just "should the new value here be 5.0? and should that be supported in the candidate selection code".

jeanconn: In testing and thorough reading of the code, it is more clear that when doing the "bright" limit checks, the effective SigErrMultiplier is always 1 and the star mag error is handled as at least 0.1 mags.  Thus the MagLimit here can be a bit brighter than 5.2 to allow stars in later states as bright *as* 5.2 .

proseco/guide.py
```diff
@@ -355,7 +355,7 @@ def search_stage(self, stage):
           {"Stage": 1,
             "SigErrMultiplier": 3,
             "ASPQ1Lim": 0,
-            "MagLimit": [5.9, 10.2],
+            "MagLimit": [5.6, 10.2],
```
jeanconn, line 358: This isn't a 0.6 shift because the example in the comment here was wrong.

PR #370, opened 2021-12-15
Exclude guide star if it overlaps with brighter/better guide star

jeanconn:
## Description

Exclude a guide star if it overlaps in pixel space with brighter/better guide star.

This PR adds a guide star selection check that guide stars are not within 12 pixels.  The PEA will drop a readout window if it actually overlaps another during the current cycle; this check should preclude star selection that would allow that.  If one or both of the close guide stars in a pair are included with include_ids(_guide) they will still be included.

This was noticed in obsids 45890 and 45884 in NOV2921A.

## Testing

- [x] Functional testing - doesn't select the overlap stars in NOV2921A

### DIFF
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -290,7 +290,7 @@ class GuideTable(ACACatalogTable):
     def exclude_overlaps(self, stage_cands):
         """
         Review the stars selected at any stage and exclude stars that overlap in
-        tracking space with another. Overlap is defined as 60 arcsecs in Y and Z.
+        tracking space with another. Overlap is defined as being within 12 pixels.
         """
         self.log(f'Checking for guide star overlap in stage-selected stars')
         nok = np.zeros(len(stage_cands)).astype(bool)
@@ -306,16 +306,19 @@ class GuideTable(ACACatalogTable):
                 # Check and exclude a guide star only if it would spoil a lower index (better) star.
                 if idx <= jdx:
                     continue
-                dy = other_star['yang'] - star['yang']
-                dz = other_star['zang'] - star['zang']
-                if np.abs(dy) < 60 and np.abs(dz) < 60:
+                drow = other_star['row'] - star['row']
+                dcol = other_star['col'] - star['col']
+                if np.abs(drow) <= 12 and np.abs(dcol) <= 12:
                     self.log(
-                        f"Rejected star {star['id']} as has track overlap with {other_star['id']}")
+                        f"Rejecting star {star['id']} with track overlap (12 pixels) "
+                        f"with star {other_star['id']}")
                     self.reject(
                         {'id': star['id'],
                          'type': 'overlap',
                          'stage': 0,
-                         'text': f'Cand {star["id"]} has track overlap with {other_star["id"]}'})
+                         'text':
+                         f'Cand {star["id"]} has track overlap (12 pixels) '
+                         f'with star {other_star["id"]}'})
                     nok[idx] = True
         return stage_cands[~nok]
 
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -296,6 +296,9 @@ class GuideTable(ACACatalogTable):
         nok = np.zeros(len(stage_cands)).astype(bool)
         for idx, star in enumerate(stage_cands):
             for jdx, other_star in enumerate(stage_cands):
+
+                # The stage_cands are supplied in the order of preference (currently by mag)
+                # Check and exclude a guide star only if it would spoil a lower index (better) star.
                 if idx <= jdx:
                     continue
                 dy = other_star['yang'] - star['yang']
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -295,6 +295,11 @@ class GuideTable(ACACatalogTable):
         self.log(f'Checking for guide star overlap in stage-selected stars')
         nok = np.zeros(len(stage_cands)).astype(bool)
         for idx, star in enumerate(stage_cands):
+
+            # If the star was manually-selected, don't bother checking to possibly exclude it.
+            if star['id'] in self.include_ids:
+                continue
+
             for jdx, other_star in enumerate(stage_cands):
 
                 # The stage_cands are supplied in the order of preference (currently by mag)
@@ -309,6 +314,7 @@ class GuideTable(ACACatalogTable):
                     self.reject(
                         {'id': star['id'],
                          'type': 'overlap',
+                         'stage': 0,
                          'text': f'Cand {star["id"]} has track overlap with {other_star["id"]}'})
                     nok[idx] = True
         return stage_cands[~nok]
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -292,7 +292,7 @@ class GuideTable(ACACatalogTable):
         Review the stars selected at any stage and exclude stars that overlap in
         tracking space with another. Overlap is defined as being within 12 pixels.
         """
-        self.log(f'Checking for guide star overlap in stage-selected stars')
+        self.log('Checking for guide star overlap in stage-selected stars')
         nok = np.zeros(len(stage_cands)).astype(bool)
         for idx, star in enumerate(stage_cands):
 
--- a/proseco/guide.py
+++ b/proseco/guide.py
@@ -281,10 +281,34 @@ class GuideTable(ACACatalogTable):
         self.log('Done with search stages')
         stage_cands = cand_guides[cand_guides['stage'] != -1]
         stage_cands.sort(['stage', 'mag'])
+        stage_cands = self.exclude_overlaps(stage_cands)
         guides = self.select_catalog(stage_cands[0:n_guide + GUIDE.surplus_stars])
         if len(guides) < self.n_guide:
             self.log(f'Could not find {self.n_guide} candidates after all search stages')
         return guides
+
+    def exclude_overlaps(self, stage_cands):
+        """
+        Review the stars selected at any stage and exclude stars that overlap in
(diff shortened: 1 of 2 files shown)


proseco/guide.py
```diff
@@ -281,11 +281,35 @@ def run_search_stages(self):
         self.log('Done with search stages')
         stage_cands = cand_guides[cand_guides['stage'] != -1]
         stage_cands.sort(['stage', 'mag'])
+        stage_cands = self.exclude_overlaps(stage_cands)
```
jeanconn, line 284: I went with a strategy of doing this at the end so there was a prioritized list of selectable stars.

proseco/guide.py
```diff
@@ -281,11 +281,35 @@ def run_search_stages(self):
...
+    def exclude_overlaps(self, stage_cands):
+        """
+        Review the stars selected at any stage and exclude stars that overlap in
+        tracking space with another. Overlap is defined as 60 arcsecs in Y and Z.
+        """
+        self.log(f'Checking for guide star overlap in stage-selected stars')
+        nok = np.zeros(len(stage_cands)).astype(bool)
+        for idx, star in enumerate(stage_cands):
+            for jdx, other_star in enumerate(stage_cands):
+                if idx <= jdx:
```
jeanconn, line 299: This could use a comment.  Since these should be in priority order by mag, I just went with the strategy of excluding the fainter star in a pair.

jeanconn: I added a comment.  There are a bunch of ways to make the loop more clear but I didn't care about optimizing.

jeanconn:
```

In [1]: import proseco                                                                                

In [2]: import pickle                                                                                 

In [3]: import gzip                                                                                   

In [4]: acas = pickle.load(gzip.open('/data/mpcrit1/mplogs/2021/NOV2921/oflsa/output/NOV2921A_proseco.
   ...: pkl.gz'))                                                                                     

In [5]: args = acas[45890].call_args.copy()                                                                                                                 

In [6]: cat = proseco.get_aca_catalog(**args)

In [7]: cat                                                                                                                                                 
Out[7]: 
<ACATable length=9>
    0     1 191765688  BOT  8x8   0.981    6.83    8.33  1947.21  1921.32    28     1   160
    1     2 191759008  BOT  8x8   0.978    7.02    8.52  2411.72  -181.47    28     1   160
    2     3 191766032  BOT  8x8   0.979    8.13    9.63    75.61 -1651.00    28     1   160
    3     4 191758776  BOT  8x8   0.973    8.63   10.13   422.49  -100.83    28     1   160
    4     5 191759456  BOT  8x8   0.963    9.01   10.51   428.90 -2444.05    28     1   160
    5     6 191765024  BOT  8x8   0.963    9.08   10.58  -456.20 -2064.06    28     1   160
    6     7 191765576  BOT  8x8   0.815    9.66   11.16  -925.24  1743.10    24     1   140
    7     8 191764552  GUI  8x8   0.000   10.02   11.20  1945.95  2019.31     1     1    25
    7     9 191758592  ACQ  8x8   0.664    9.37   10.87  2472.67 -1144.07    28     1   160

In [8]: 191762344 in cat['id']                                                                                                                              
Out[8]: False

In [9]: 191765024 in cat['id']                                                                                                                              
Out[9]: True


```

proseco/guide.py
jeanconn: This still only considers guide stars (not fids or mons).

proseco/guide.py
```diff
@@ -281,11 +281,39 @@ def run_search_stages(self):
...
+
+                # The stage_cands are supplied in the order of preference (currently by mag)
+                # Check and exclude a guide star only if it would spoil a lower index (better) star.
+                if idx <= jdx:
+                    continue
+                dy = other_star['yang'] - star['yang']
+                dz = other_star['zang'] - star['zang']
+                if np.abs(dy) < 60 and np.abs(dz) < 60:
+                    self.log(
+                        f"Rejected star {star['id']} as has track overlap with {other_star['id']}")
```
taldcroft, line 308:
```suggestion
                        f"Rejecting star {star['id']} with track overlap (12 pixels) of selected star {other_star['id']}")
```

proseco/guide.py
```diff
@@ -281,11 +281,39 @@ def run_search_stages(self):
...
+                    self.reject(
+                        {'id': star['id'],
+                         'type': 'overlap',
+                         'stage': 0,
+                         'text': f'Cand {star["id"]} has track overlap with {other_star["id"]}'})
```
taldcroft, line 313:
```suggestion
                         'text': f'Cand {star["id"]} has track overlap (12 pixels) of selected star {other_star["id"]}'})
```

proseco/tests/test_guide.py
```diff
@@ -316,15 +316,31 @@ def test_check_spoiler_cases():
...
             selected = get_guide_catalog(**STD_INFO, stars=stars, dark=dark)
             spoiled.append(1 if (1 not in selected['id']) else 0)
     spoiled = np.array(spoiled).reshape(-1, len(drcs)).tolist()
     #                    0  2  4  6  8 10 12 pixels
     expected_spoiled = [[1, 1, 1, 1, 0, 0, 0],  # dmag = 3
                         [1, 1, 1, 1, 0, 0, 0],  # dmag = 5
                         [1, 1, 1, 1, 0, 0, 0]]  # dmag = 7
+    assert spoiled == expected_spoiled
 
+    # Add a brighter "spoiler" that is a good guide star and confirm the test for
```
taldcroft, line 329: Put this into a new test function, i.e. try to make tests conceptually narrow. If worried about repeating yourself in setup this can be handled with pytest fixtures, though in this case it is easier to just repeat the necessary setup.

proseco/tests/test_guide.py
```diff
@@ -296,7 +296,7 @@ def test_check_spoiler_cases():
             stars.add_fake_star(row=r, col=c, mag=mag0, id=1, ASPQ1=1)
             # Add a "spoiling" star and move it from center past edge through
             # the drcs
-            stars.add_fake_star(row=r + drc, col=c, mag=mag0 + dmag, id=2, ASPQ1=0)
+            stars.add_fake_star(row=r + drc, col=c, mag=mag0 + dmag, id=2, ASPQ1=30)
```
taldcroft, line 299: Why are you changing this existing test?

proseco/tests/test_guide.py
```diff
@@ -316,15 +316,31 @@ def test_check_spoiler_cases():
...
+    for drc in drcs:
+        r = 10
+        c = 10
+        stars = StarsTable.empty()
+        stars.add_fake_star(row=r, col=c, mag=mag0, id=1, ASPQ1=0)
+        # Add a brighter spoiling star
+        stars.add_fake_star(row=r + drc, col=c + drc, mag=mag0 - 0.5, id=2, ASPQ1=0)
+        selected = get_guide_catalog(**STD_INFO, stars=stars, dark=dark)
+        spoiled.append(1 if (1 not in selected['id']) else 0)
+    spoiled = np.array(spoiled).reshape(-1, len(drcs)).tolist()
```
taldcroft, line 341: This looks like a complicated way to turn a `list` into `list`.  I guess inherited from above where there were multiple dmags.

proseco/guide.py
taldcroft: Fids check for stars bright enough to be a guide star with regular spoiler checks. And we can check in sparkles in case that isn't enough. Algorithmically doing everything in proseco does get a bit tricky.

proseco/guide.py
taldcroft: (I used Rejecting instead of Rejected because the latter is grammatically ambiguous).

proseco/tests/test_guide.py
jeanconn: So that the id=2 star can't be used as a guide star, thus hitting our new check.

proseco/tests/test_guide.py
jeanconn: OK.  I thought the "spoiling" still was conceptually narrow, but no trouble to break out.

jeanconn:
@taldcroft What do you want the behavior to be if one or both of the pair of stars are in include_ids_guide?  Exclude as needed or just warn in sparkles?

taldcroft:
> What do you want the behavior to be if one or both of the pair of stars are in include_ids_guide? Exclude as needed or just warn in sparkles?

If one of the pair of stars is in `include_ids_guide` then it should be the selected star. If both are included in the list then they should both be selected. That's the API for `include_ids_guide` (do what I say, not what you think is best for me).

jeanconn:
Right, there's some precedent for not allowing an "unselectable" star even if in the include_ids_guide but fine to handle this like the other "bad idea but not precluded" cases.

proseco/guide.py
```diff
@@ -281,11 +281,47 @@ def run_search_stages(self):
...
+
+            for jdx, other_star in enumerate(stage_cands):
+
+                # The stage_cands are supplied in the order of preference (currently by mag)
+                # Check and exclude a guide star only if it would spoil a lower index (better) star.
+                if idx <= jdx:
+                    continue
+                drow = other_star['row'] - star['row']
+                dcol = other_star['col'] - star['col']
+                if np.abs(drow) <= 12 and np.abs(dcol) <= 12:
```
jeanconn, line 311: If the check should really be 12 pixels (instead of 60 arcsecs) I changed it to 12 pixels.

proseco/tests/test_guide.py
jeanconn: Though probably should have changed the test by making the fake star un-selectable via class or some such that is not a stage-selection-influencing piece.

taldcroft (approved):
Looks good!

PR #371, opened 2022-01-20
Prevent matplotlib warning

javierggt:
## Description

This PR prevents this warning that appeared int sot/skare3/pull/755
```
FixedFormatter should only be used together with FixedLocator
```

Maybe there is a better way?

## Testing

- [x] Passes unit tests on MacOS (no warnings emitted)
- [x] Functional testing

### Functional testing

TA ran star selection for obsid 8008 using this PR and visually inspected the acq report and noticed no problems with the imposter plots (and all plots in general). Also did a side-by-side comparison with the same output from current master and saw no diffs.

### DIFF
--- a/proseco/report_acq.py
+++ b/proseco/report_acq.py
@@ -7,6 +7,7 @@ from pathlib import Path
 import matplotlib
 matplotlib.use('agg')  # noqa
 from matplotlib import patches
+from matplotlib.ticker import FixedLocator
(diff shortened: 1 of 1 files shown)


taldcroft (approved):
I did functional testing and ran unit tests (documented in the description) and it looks good. I'm not enough of a matplotlib expert to suggest a better way but this looks good to me.

PR #376, opened 2022-08-03
Adjust acq selection to limit search hits to < 50

taldcroft:
## Description

This implements the constraint imposed by the contour plot in https://occweb.cfa.harvard.edu/twiki/bin/view/Aspect/PeaMaxMagTesting.

## Interface impacts

No API change in `get_aca_catalog` or other functional interfaces.

The only update of note is that setting the environment variable `PROSECO_IGNORE_MAXMAGS_CONSTRAINTS` to any value will cause proseco to ignore the search hits limit in this PR. This is included for testing.

### Unit tests

- [x] Mac

Independent check of unit tests by [REVIEWER NAME]
- [ ] [PLATFORM]:

### Functional tests

- [x] Reselect catalogs for 4 recent weekly loads at CCD temperature of -4.0 C and review diffs. This notebook is included in this PR. See [validate-maxmags-recent-loads.ipynb](https://nbviewer.org/urls/cxc.cfa.harvard.edu/mta/ASPECT/proseco/validate-maxmags-recent-loads.ipynb)


proseco/acq.py
```diff
@@ -27,6 +33,86 @@
...
+    :param mag: star mag (float)
+    :param mag_err: star mag error (float)
+    :param box_sizes: ndarray of box sizes (float, arcsec)
+    :param t_ccd: CCD temperature (float, C)
+    :returns: ndarray of box sizes (float, arcsec)
+    """
+    maxmag_min = mag + np.clip(3 * mag_err, ACA.min_delta_maxmag, ACA.max_delta_maxmag)
+
+    # Hard limit of ACA.max_maxmag (11.2) from operational change made in 2019.
+    # We always accept a maxmax of 11.2 regardless of star mag / mag_err.
```
jeanconn, line 102: Typo in comment and I'm not really sure what this comment means with regard to accepting (accepting the magmax or accepting the box size).

proseco/acq.py
```diff
@@ -27,6 +33,86 @@
...
+    maxmag_min = maxmag_min.clip(None, ACA.max_maxmag)
+
+    ok = [maxmag_min <= get_maxmag(box_size, t_ccd) for box_size in box_sizes]
+    out: np.ndarray = box_sizes[ok]
+
+    # Always allow at least the smallest box size. This situation will be
```
jeanconn, line 108: Why do we need to allow at least the smallest box size and how is this flagged?

taldcroft: Because it will be better to have an acq star with a 60 arcsec box (where the actual acq prob is less than the model) than nothing at all. It will be flagged in sparkles which will check the commanded maxmag vs allowed given T_ccd and box size.

jeanconn: Sounds fine but where should we track the lien?

taldcroft: Sparkles issue.

jeanconn:
I added related issue https://github.com/sot/sparkles/issues/176 which will need the maxmag data.

jeanconn (approved):
I might change the PR title to something like "Adjust acq selection to limit search hits to < 50" as the allowed box candidates and the maxmag is adjusted as needed to make this work.  Not a required change.

taldcroft:
Thanks for the review @jeanconn !

jeanconn:
I note with this merged that we're seeing a "fail" in ska_testr sparkles just based on the new warning:
```
  /export/jgonzale/github-workflows/miniconda3/envs/ska3-masters/lib/python3.8/site-packages/proseco/acq.py:70: UserWarning: Clipping t_ccd=10 to 0.0 for interpolating MAXMAGs table
    warnings.warn(f"Clipping {t_ccd=} to 0.0 for interpolating MAXMAGs table")
```
Since we plan a release of sparkles to go with this anyway, this is benign.

PR #379, opened 2022-09-19
Change default OR guide and acq image size to 8x8 pixels

### LINKED ISSUE #378: Make 8x8 pixel image readout be default for OR's
Per agreement of SS&AWG we should update proseco to default to using 8x8 pixel images for all guide stars. In particular this changes the default for OR catalogs from 6x6 pixels to 8x8 pixels.

This is targeted for inclusion in MATLAB 2022_060.

taldcroft:
## Description

Per agreement of SS&AWG this PR updates proseco to default to using 8x8 pixel images for all guide stars. In particular this changes the default for OR catalogs from 6x6 pixels to 8x8 pixels.

This is targeted for inclusion in MATLAB 2022_060.

The new default can be globally overridden with an environment variable `PROSECO_OR_IMAGE_SIZE` which can be set to one of `4`, `6`, or `8`.

Closes #378

## Interface impacts

Changes the default guide star image size for OR's (observations with fids) from 6x6 to 8x8.

### Unit tests

- [x] Mac

- [X] Linux

### Functional tests

No functional testing.

### DIFF
--- a/proseco/catalog.py
+++ b/proseco/catalog.py
@@ -717,7 +717,7 @@ def merge_cats(fids=None, guides=None, acqs=None, mons=None):
 
     if len(acqs) > 0:
         # TODO: move these into acq.py where possible
-        img_size = get_img_size(len(fids))
+        img_size = guides.img_size or get_img_size(len(fids))
         acqs["type"] = "ACQ"
 
         # Set maxmag for acqs. Start w/ legacy version corresponding to behavior
--- a/proseco/__init__.py
+++ b/proseco/__init__.py
@@ -75,7 +75,8 @@ def get_aca_catalog(*args, **kwargs):
     :param include_ids_guide: list of AGASC IDs of stars to include in guide catalog
     :param exclude_ids_guide: list of AGASC IDs of stars to exclude from guide catalog
     :param img_size_guide: readout window size for guide stars (6, 8, or ``None``).
-                           For default value ``None`` use 8 for no fids, 6 for fids.
+        Default value: 8 for ER's; PROSECO_OR_IMAGE_SIZE environment value if
+        if set, otherwise 8.
     :param optimize: optimize star catalog after initial selection (default=True)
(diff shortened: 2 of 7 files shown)


proseco/catalog.py
```diff
@@ -717,7 +717,7 @@ def merge_cats(fids=None, guides=None, acqs=None, mons=None):
 
     if len(acqs) > 0:
         # TODO: move these into acq.py where possible
-        img_size = get_img_size(len(fids))
+        img_size = guides.img_size or get_img_size(len(fids))
```
jeanconn, line 720: Still weird that we care here for the display (printing the unused image size of acqs).

taldcroft: Well this parameter actually gets to the command loads (i.e. it's in the DOT) and is uplinked to the spacecraft. So technically it is not just printing an unused value. And I'd have to double-check, but I think that it gets through to the ACA search command and might show up in a couple readouts before the OBC sends the guide search command.

jeanconn: I thought the acq box readouts were always 8x8 but I was focused more on the PEA side so that's a good question.

taldcroft: In obsid 24771 there is an ACQ star with 6x6 in starcheck, and I see IMGSZ11= 1 in the backstop file. So I can't say what is getting sent to the PEA, but I'm pretty sure that 6x6 is in the command loads and is going up to the spacecraft. From that perspective it matters.

jeanconn: OK.  Good to have this talk.  The image size for acquisition stars just isn't something I've considered or that we've talked about in this transition... right down to this PR title.

taldcroft: Good point about the PR title, fixed.

taldcroft: I looked at image telemetry for obsid 8008 and confirmed that the 2-3 readouts of acq-only images were 6x6 before the fids were found. So the size value for ACQ makes it all the way to the PEA.

PR #385, opened 2023-08-15
Add test of optional ACA limits

taldcroft:
## Description

This adds a unit test that verifiers the functionality in #384.

As a small out-of-scope change, move to `matplotlib.pyplot.switch_backend` to avoid the linting problem with imports after code.

## Interface impacts

None.

### Unit tests

- [x] Mac

- [x] Linux

### Functional tests

No functional testing.

### DIFF
--- a/proseco/tests/test_catalog.py
+++ b/proseco/tests/test_catalog.py
@@ -1,7 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import copy
 import os
-import pickle  # noqa
+import pickle
 from pathlib import Path
 
 import agasc
@@ -17,7 +17,7 @@ from ..core import ACACatalogTable, StarsTable, includes_for_obsid
 from ..fid import FidTable, get_fid_catalog
 from .test_common import DARK40, OBS_INFO, STD_INFO, mod_std_info
 
-# Switch the matplotlib backend to Agg
+# Ensure all plotting is to a non-interactive backend
 matplotlib.pyplot.switch_backend("agg")
 
 HAS_SC_ARCHIVE = Path(mica.starcheck.starcheck.FILES["data_root"]).exists()
--- a/proseco/tests/test_catalog.py
+++ b/proseco/tests/test_catalog.py
@@ -1,25 +1,24 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import copy
 import os
-
-import matplotlib
-from Quaternion import Quat
-
-matplotlib.use("agg")  # noqa
-
-import pickle
+import pickle  # noqa
 from pathlib import Path
 
 import agasc
+import matplotlib
 import mica.starcheck
 import numpy as np
 import pytest
+from Quaternion import Quat
 
 from .. import characteristics as ACA
 from ..catalog import ACATable, get_aca_catalog
 from ..core import ACACatalogTable, StarsTable, includes_for_obsid
 from ..fid import FidTable, get_fid_catalog
 from .test_common import DARK40, OBS_INFO, STD_INFO, mod_std_info
+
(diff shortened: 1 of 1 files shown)


proseco/tests/test_catalog.py
```diff
@@ -398,6 +397,52 @@ def test_big_sim_offset():
...
+        path_aca_spec = (
+            Path(repo_path_local) / "chandra_models" / "xija" / "aca" / "aca_spec.json"
+        )
+        text_aca_spec = path_aca_spec.read_text()
+        aca_spec = json.loads(text_aca_spec)
+        del aca_spec["limits"]["aacccdpt"]["planning.penalty.high"]
+        repo.git.checkout("HEAD", b="test_branch")
+        path_aca_spec.write_text(json.dumps(aca_spec, indent=4))
+        repo.git.add(path_aca_spec)
+        repo.git.commit("-m", "test commit")
```
jeanconn, line 432: This seems fine, though is it basically a proof-of-concept?  I'd I've figured if you aren't doing a detailed comparison with 3.48 results, you'd have used the sha of the chandra_models on deck without the penalty limit (https://github.com/sot/chandra_models/pull/113/commits/293926949b4059db35d7d498dd3438cb5f2ecc73)

taldcroft: This is intended as an example of testing code changes related to updates in chandra_models that are not yet available in flight Ska data. So it makes the change via code on the fly instead of relying on a repo.

The commit you mentioned is still in a branch on GitHub. Currently we don't have a process for fetching remote branches into the flight Ska chandra_models. I suppose we *could* do that, in which case your idea of just using version="2939269" would be preferable.

Maybe I'm missing something? This approach was partly motivated by a previous comment you made that the test I wrote would fail in masters testing because it relied on a not-yet-existent commit in chandra_models.

jeanconn: Right, this wasn't a criticism as much as a process question.  So yes, it sounds like this is a proof of concept and example.  IIRC we decided the sha method was acceptable for the other test because we were OK with the test failing on flight ska3 data until the sha was available in flight /proj/sot/ska/data/chandra_models and I think @javierggt  updated the masters ska data to just have the chandra_models repo with the needed sha.  If we want the test to run now without fiddling around with ska data and the edit is simple (remove the key), this seems fine.

taldcroft: Yes, I'm happy with the approach for now.

taldcroft:
@jeanconn - please approve if you are good with this.

jeanconn:
I eye-balled the new git and model fetching but haven't had a chance to do review on this yet.

taldcroft:
Ah OK, I misinterpreted the scope of "This seems fine". No problem.

proseco/tests/test_catalog.py
```diff
@@ -398,6 +397,52 @@ def test_big_sim_offset():
     assert aca.target_name == "Target Name"
 
 
+def test_optional_penalty_limit():
+    """Check that the optional penalty limit is supported correctly."""
+    try:
+        _test_optional_penalty_limit()
+    finally:
+        # Reset ACA limits characteristics that get set above
+        ACA._set_aca_limits()
```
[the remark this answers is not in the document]
taldcroft: I wasn't exactly sure how to do this with a fixture and decided this was good enough.

jeanconn (approved):
Looks good.  The business with manually running ACA._set_aca_limits() to override the proseco "cache" and lazy load for these attributes reminded me... should proseco also be using @chandra_models.chandra_models_cache to get the model?

taldcroft:
> should proseco also be using @chandra_models.chandra_models_cache to get the model?

Good point. I think something like this can work and be a good thing. Next PR...

PR #388, opened 2023-10-17
Apply temperature-dependent drift model offset to fid positions

jeanconn:
## Description

Update proseco to apply the time and temperature-dependent fid position offsets to the fid positions in proseco constructed catalogs.  This code relies on `chandra_aca.drift.get_fid_offset` and the aimpoint drift model in `chandra_models`.

## Interface impacts
Changes default behavior in proseco to apply offsets in y and z to expected fid positions via chandra_aca.drift.get_fid_offset.  Respects new environment variable PROSECO_ENABLE_FID_OFFSET .

## Testing
Requires https://github.com/sot/chandra_aca/pull/157

### Unit tests

- [x] Mac

Independent check of unit tests by @taldcroft
- [x] Mac

### Functional tests

Added a unit test.
Reselected ~900 catalogs in two aca aimpoint drift model epochs and since the last safe modes (2023:044 and 2023:047) and confirmed proseco fid positions with the PR are overall reasonable compared to observed positions and that there are small mean offests since the last safe mode (expected).  Also reviewed reselected star catalogs with critical warnings (with and without fid offsets applied) and confirmed proseco catalogs are reasonable (one new critical warning on an hrc fid light in a field with many spoiler stars).

Functional review notebook is part of the PR in the analysis directory https://github.com/sot/proseco/blob/5595c4c002814b82ccee2f20f4f084bd14850770/analysis/reselect_fid_offset.ipynb


taldcroft (commented):
I know this is a WIP but it looks really good and just needs a few changes and some testing. We should go ahead with this soon!

proseco/fid.py
```diff
@@ -472,7 +475,7 @@ def set_spoilers_score(self, fid):
                 )
 
 
-def get_fid_positions(detector, focus_offset, sim_offset):
+def get_fid_positions(detector, focus_offset, sim_offset, t_ccd_acq=None, date=None):
```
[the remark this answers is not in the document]
taldcroft: Also docstring update.

proseco/fid.py
```diff
@@ -518,4 +521,22 @@ def get_fid_positions(detector, focus_offset, sim_offset):
 
     yang, zang = np.degrees(yfid) * 3600, np.degrees(zfid) * 3600
 
+    # Apply fid offset
```
taldcroft, line 524: Comments should reflect *why* something is being done, not *what* is being done which is clear from the call to `get_fid_offset`.

jeanconn: Which I suppose also speaks to deleting comments after CoPilot.

proseco/fid.py
```diff
@@ -518,4 +521,22 @@ def get_fid_positions(detector, focus_offset, sim_offset):
...
+    enable_fid_offset_env = os.environ.get("PROSECO_ENABLE_FID_OFFSET", "True")
```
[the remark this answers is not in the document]
jeanconn: Right.  I was following the model of AGASC_SUPPLEMENT_ENABLE to some extent, figured that anything with "no need to check the value" was somewhat asking for trouble, and wasn't sure if we'd have a use case to use the opposite default (either here or via config).

taldcroft:
There is a slight complication here:
https://github.com/sot/kadi/blob/76b7d0b6b79102ab60c53df302eab1eb42c65b0e/kadi/commands/observations.py#L85

The current code will break this because `t_ccd` is not known. This supports a thought I was having independently, which is that if `t_ccd` and `date` are both `None` then that should imply not adding the offset. That also leads to a 3-state value for `PROSECO_ENABLE_FID_OFFSET`:
- "True": always apply and require that `t_ccd` and `date` are supplied
- "False": never apply
- None (undefined, which will be the default): apply if `t_ccd` and `date` are both supplied.

jeanconn:
Those kadi observations are an interesting use case.  Thanks for bringing that up.

jeanconn:
I forgot -- I need to document the functional testing in here so that's still to-do.

jeanconn:
We probably also need to handle dates < 2012 explicitly (perhaps with a warning instead of rethrowing the valueerror from chandra_aca) if we want to leave this turned on and run old schedules in matlab regression for example.

taldcroft:
Good point about the 2012 cutoff from this code in `chandra_aca.drift`:
```
        if times[0] < DateTime("2012:001:12:00:00").secs:
            raise ValueError("model is not applicable before 2012")
```
One idea is to just change the drift model to extrapolate back from 2012 with a constant via:
```
        times = times.clip(DateTime("2012:001:12:00:00").secs, None)
```

taldcroft:
One more thing, can you make the notebook names more descriptive. I'd suggest:
- `pr388-backstop-vs-proseco-star-positions.ipynb`
- `pr388-validate-drift-model-fid-positions.ipynb`

proseco/fid.py
```diff
@@ -518,4 +536,26 @@ def get_fid_positions(detector, focus_offset, sim_offset):
...
+            raise ValueError(
+                "t_ccd_acq and date must be provided if PROSECO_ENABLE_FID_OFFSET is 'True'"
+            )
+
+        # Apply fid offset if t_ccd and date are provided
+        # (this is an "if" the case when enable_fid_offset_env is None)
+        if t_ccd is not None and date is not None:
+            dy, dz = get_fid_offset(date, t_ccd)
+            yang += dy
+            zang += dz
```
taldcroft, line 559: You can reduce the complexity like so, to the point that code comments are not even needed:
```suggestion
    has_tccd_and_date = t_ccd is not None and date is not None

    if enable_fid_offset_env == "True" and not has_tccd_and_date:
        raise ValueError(
            "t_ccd_acq and date must be provided if PROSECO_ENABLE_FID_OFFSET is 'True'"
        )

    if has_tccd_and_date and enable_fid_offset_env != "False":
        dy, dz = get_fid_offset(date, t_ccd)
        yang += dy
        zang += dz
```

proseco/tests/conftest.py
```diff
@@ -7,6 +7,11 @@ def use_fixed_chandra_models(monkeypatch):
     monkeypatch.setenv("CHANDRA_MODELS_DEFAULT_VERSION", "3.48")
 
 
+@pytest.fixture(autouse=True)
```
taldcroft, line 10: Can you try without `autouse=True` and see how many tests fail? I'm getting less keen on the idea of our tests defaulting to a non-flight behavior for everything.

proseco/tests/test_fid.py
```diff
@@ -10,7 +10,9 @@
 from ..fid import get_fid_catalog, get_fid_positions
 from .test_common import DARK40, OBS_INFO, STD_INFO, mod_std_info
 
-# Reference fid positions for spoiling tests
+# Reference fid positions for spoiling tests.  Because this is done outside
+# a test the fixture to generally disable fid drift is not in effect.
+os.environ["PROSECO_ENABLE_FID_OFFSET"] = "False"
```
taldcroft, line 15: Use the [temp_env_var](https://sot.github.io/ska_helpers/#ska_helpers.utils.temp_env_var) context manager here.

I'm not happy about the `agasc.SUPPLEMENT_ENABLED_ENV` line below either, but that is for another day.

proseco/tests/test_fid.py
```diff
@@ -49,6 +51,56 @@ def test_get_fid_position():
     assert np.allclose(zang[fidset], [-468, -460, 561], rtol=0, atol=1.1)
 
 
+def test_get_fid_pos_with_offsets(monkeypatch):
+    # Confirm that if env var is set to 'False' then no offset is applied.
+    # Env var is set to false in the fixture.
```
taldcroft, line 56: Be explicit with `monkeypatch.setenv("PROSECO_ENABLE_FID_OFFSET", "False")`.

proseco/tests/test_fid.py
```diff
@@ -49,6 +51,56 @@ def test_get_fid_position():
...
+    assert np.allclose(zang1 - zang2, -9.63, rtol=0, atol=0.1)
+
+    # Confirm that if env var is not set and t_ccd and date are specified, then
+    # offset is applied.
+    monkeypatch.delenv("PROSECO_ENABLE_FID_OFFSET", raising=False)
+    yang3, zang3 = get_fid_positions(
+        "ACIS-S", focus_offset=0.0, sim_offset=0.0, date="2023:235", t_ccd=-13.65
+    )
+    assert np.allclose(yang2, yang3, rtol=0, atol=0.1)
+    assert np.allclose(zang2, zang3, rtol=0, atol=0.1)
```
taldcroft, line 75: The results should be identical, no?

proseco/tests/test_fid.py
```diff
@@ -49,6 +51,56 @@ def test_get_fid_position():
...
+    # Confirm that if env var is not set and t_ccd and date not specified, then
+    # no offset is applied.
+    monkeypatch.delenv("PROSECO_ENABLE_FID_OFFSET", raising=False)
+    yang4, zang4 = get_fid_positions(
+        "ACIS-S",
+        focus_offset=0.0,
+        sim_offset=0.0,
+    )
+    assert np.allclose(yang1, yang4, rtol=0, atol=0.1)
+    assert np.allclose(zang1, zang4, rtol=0, atol=0.1)
```
taldcroft, line 86: Identical.

proseco/tests/test_fid.py
```diff
@@ -49,6 +51,56 @@ def test_get_fid_position():
...
+
+    # Confirm if env var is set to 'True' and t_ccd and date not specified, then
```
taldcroft, line 88: Put the error handling tests in a separate test function. In general tests should always be as small and atomic as reasonable so that one failure does not block running other tests.

proseco/tests/test_fid.py
```diff
@@ -49,6 +51,56 @@ def test_get_fid_position():
...
+    # Confirm an error if env var set to not-allowed value
+    monkeypatch.setenv("PROSECO_ENABLE_FID_OFFSET", "foo")
+    with pytest.raises(
+        ValueError, match='env var must be either "True", "False", or None,'
+    ):
+        get_fid_positions(
+            "ACIS-S", focus_offset=0.0, sim_offset=0.0, t_ccd=-13.65, date="2023:235"
+        )
+
+
```
taldcroft, line 103: We need an "integration regression test" with calling `get_aca_catalog`.

proseco/tests/test_fid.py
jeanconn: I wasn't sure how much cleanup to do while I was in here, but it seems like both of these should just move into fixtures.

proseco/tests/conftest.py
jeanconn: It is a subtlety I'd missed that basically all the tests pass without this set now that we've moved to the tri-state on PROSECO_ENABLE_FID_OFFSET and our standard get_aca_catalog test setup doesn't actually set t_ccd_acq https://github.com/sot/proseco/blob/85c79bef6f1b7ef9594706d23743a1b947b3a08f/proseco/tests/test_common.py#L14 .  I think we should probably update our standard regression tests to call get_aca_catalog the way it is used in operations.

proseco/fid.py
```diff
@@ -122,7 +126,7 @@ def acqs(self, val):
 
     @property
     def t_ccd(self):
-        # For fids use the guide CCD temperature
+        # For fids use the guide CCD temperature to scale the dark map
```
jeanconn, line 129: I'm pretty sure that's how t_ccd uses, but the logic in here is a little twisty.

proseco/fid.py
```diff
@@ -94,6 +97,7 @@ class FidTable(ACACatalogTable):
         "detector",
         "sim_offset",
         "focus_offset",
+        "t_ccd_acq",
```
jeanconn, line 100: I don't think it is controversial to just require the acquisition temperature too if that is the best temperature to use for the fid offsets.

proseco/fid.py
```diff
@@ -128,6 +131,7 @@ def t_ccd(self):
     @t_ccd.setter
     def t_ccd(self, value):
         self.t_ccd_guide = value
+        self.t_ccd_acq = value
```
jeanconn, line 134: In more detailed review and testing, _this_ was the change needed so that "get_fid_catalog" will mostly do the right thing if *just* a t_ccd is supplied.  In that case the temperature is assigned to both t_ccd_guide and t_ccd_acq and fid offsets will be applied based on t_ccd_acq by default.  This was already the default behavior for get_aca_catalog, but not for get_fid_catalog.

taldcroft: Seems reasonable.

proseco/tests/test_common.py
```diff
@@ -11,7 +11,8 @@
     date="2018:001",
     n_guide=5,
     n_fid=3,
-    t_ccd=-11,
+    t_ccd_acq=-11,
+    t_ccd_guide=-11,
```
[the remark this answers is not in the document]
taldcroft: Good idea for another day.

proseco/core.py
```diff
@@ -1487,7 +1487,8 @@ def add_fake_stars_from_fid(
                 detector=detector,
                 sim_offset=sim_offset,
                 focus_offset=0,
-                t_ccd=-10.0,
+                t_ccd_guide=-10.0,
+                t_ccd_acq=-10.0,
```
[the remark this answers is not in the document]
taldcroft: Can reverting this count as the test of changing the `t_ccd` setter?

taldcroft:
Is this ready again for review?  BTW I put this as Squash and Merge, unless you are adamant that all 36 commits should go in the permanent history.

proseco/core.py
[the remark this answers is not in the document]
jeanconn: I've just un-done this (makes sense at this point), though did your comment mean that you wanted me to remove some of my gratuitous t_ccd tests?

jeanconn:
Sure.  I just made that change to add_fake_stars_from_fid and reran unit tests (did not re-run functional/notebook).  This is ready for review again.  And sure I don't care if we keep all the commits if this is approved.

taldcroft (approved):
@jeanconn - I pushed a final commit with some cleanup. I re-reviewed the key validation notebook and code. It all looks good now, thanks for this important fix!

proseco/tests/test_catalog.py
```diff
@@ -568,10 +565,84 @@ def test_call_args_attr():
...
 
 
+def test_t_ccd_attr():
+    """
+    Test that a t_ccd kwarg is set to all of the ACACatalogTable attrs
+    """
+
+    aca_args = STD_INFO.copy()
+    for kw in ["t_ccd", "t_ccd_acq", "t_ccd_guide"]:
+        if kw in aca_args:
```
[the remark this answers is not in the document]
taldcroft: And if it ever comes up again, we could update `mod_info()` (or whatever is that function) to allow for deleting kwargs via `mod_info(delete_kwargs=("t_ccd", "t_ccd_acq", "t_ccd_guide"))`.

PR #407, opened 2025-07-11
Add function to split search hit regions

taldcroft:
## Description

This fixes an issue where `get_imposter_stars` could miss imposters at a warm temperature and faint mag.
```
>>> dark = dark_cal.get_dark_cal_image(date='2025:190', t_ccd_ref=-2.0).astype(np.float64)
>>> # Using a local version of get_imposter_stars which has slightly different outputs
>>> imposters = get_imposter_stars(dark[:60, :60], maxmag=11.2, mag_limit=10.25)
>>> imposters.pprint_all()  # After seeing float format to .2f
n_vals   yang    zang   row  col  img_sum  mag  mag_err
------ ------- -------- ---- ---- ------- ----- -------
    24 2520.72 -2523.51 -506 -506 9354.07  9.70    0.11
   579 2398.74 -2397.03 -481 -480 8613.77  9.79    0.12  <== Mega search hit region
     1 2287.31 -2496.52 -459 -501 6107.14 10.16    0.15
```

This PR adds a new function to split search hit regions into boxes no larger than 8x8. This simple strategy is not perfect, but it is much better than the current situation in which one mega region gets a single imposter with the mag of the 6x6 region at the center. The new code will generate many imposters that effectively block selecting the candidate star in that region.

### Unit tests

- [x] Mac (with new unit test)
```
(ska3) ➜  proseco git:(acq-split-label-search-hit-regions) git rev-parse --short HEAD                             
8791b0c
(ska3) ➜  proseco git:(acq-split-label-search-hit-regions) pytest 
=============================================== test session starts ================================================
platform darwin -- Python 3.12.8, pytest-8.3.4, pluggy-1.5.0
rootdir: /Users/aldcroft/git
configfile: pytest.ini
plugins: anyio-4.7.0, timeout-2.3.1
collected 172 items                                                                                                

proseco/tests/test_acq.py ........................................                                           [ 23%]
proseco/tests/test_catalog.py ..........................................                                     [ 47%]
proseco/tests/test_core.py ............................                                                      [ 63%]
proseco/tests/test_diff.py ......                                                                            [ 67%]
proseco/tests/test_fid.py ...............                                                                    [ 76%]
proseco/tests/test_guide.py .....................................                                            [ 97%]
proseco/tests/test_mon_full_cat.py ....                                                                      [100%]

=============================================== 172 passed in 27.45s ===============================================
```

Independent check of unit tests by [REVIEWER NAME]
- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/proseco/acq.py
+++ b/proseco/acq.py
@@ -1343,13 +1343,15 @@ def get_imposter_stars(
     :param star_col: col of acq star (float)
     :param thresh: PEA search hit threshold for a 2x2 block (e-/sec)
     :param maxmag: Max mag (alternate way to specify search hit ``thresh``)
-    :param box_size: box size (arcsec)
+    :param box_size: float or ACABox box size (arcsec)
     :param bgd: assumed flat background (float, e-/sec)
     :param mag_limit: Max mag for imposter (using 6x6 readout)
     :param test: hook for convenience in algorithm testing
 
     :returns: numpy structured array of imposter stars
     """
+    box_size = ACABox(box_size)
+
     # Convert row/col to array index coords unless testing.
     rc_off = 0 if test else 512
     acq_row = int(star_row + rc_off)
@@ -1393,6 +1395,11 @@ def get_imposter_stars(
     # contiguous regions above ``thresh`` labeled with a unique index.
     # This is a one-line way of doing the PEA merging process, roughly.
     dc_labeled, n_hits = ndimage.label(dc2x2 > thresh)
+
+    # If no hits just return empty list
+    if n_hits == 0:
+        return []
+
     if ACQ.search_hit_max_size is not None:
         dc_labeled, n_hits = split_labeled_regions(
             dc_labeled, max_size=ACQ.search_hit_max_size
@@ -1401,10 +1408,6 @@ def get_imposter_stars(
         # Print the entire dc_labeled array without clipping
         with np.printoptions(threshold=100000, linewidth=200):
             print(dc_labeled)
-
-    # If no hits just return empty list
-    if n_hits == 0:
-        return []
 
     outs = []
     for idx in range(n_hits):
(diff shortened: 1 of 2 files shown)


jeanconn:
While I understand the justification here, and the unit testing looks good,  I would expect to see something like a notebook showing that we get similar star selection with this PR for cold temperatures and for hot temperatures there are, in aggregate, changes to acq catalogs, and then a walk-through showing that at least one acquisition catalog has a more sane rejection or selection around an imposter region.  Do you want me to go and do that?

proseco/acq.py
```diff
@@ -1346,8 +1393,14 @@ def get_imposter_stars(
     # contiguous regions above ``thresh`` labeled with a unique index.
     # This is a one-line way of doing the PEA merging process, roughly.
     dc_labeled, n_hits = ndimage.label(dc2x2 > thresh)
+    if ACQ.search_hit_max_size is not None:
+        dc_labeled, n_hits = split_labeled_regions(
+            dc_labeled, max_size=ACQ.search_hit_max_size
```
jeanconn, line 1398: This is throwing "ValueError: zero-size array to reduction operation maximum which has no identity" on my first toy test at `slices = ndimage.find_objects(dc_labeled)`, so unless you want to handle the empty case in split_labeled_regions, I think the "if n_hits == 0: return" code should probably come before split_labeled_regions.

taldcroft: Can you provide a reproducible example? I tried the following and it worked as expected.
```
dc_labeled = np.array([[0, 0], [0, 0]])
pracq.split_labeled_regions(dc_labeled)
```
I agree that the `if n_hits == 0:` check could be moved up for efficiency.

jeanconn: It looks like my original issue was that I both wasn't sure about the domain of "star_row" and "star_col" and I'm a bit flummoxed about why "test" changes the offset by 512 pixels.  It looks like now I'm only hitting ValueError if I'm out of bounds.

```
dark = mica.archive.aca_dark.get_dark_cal_image("2025:001", t_ccd_ref=-10)
dat = Table(get_imposter_stars(dark=dark, star_row=200, star_col=2000, box_size=(1000,1000)))
```

